PR #83, opened 2020-04-06
show documentation warnings in GitHub actions

ntessore:
## Description
This adds a very naive detection for documentation build errors. Since the error and warning messages are a mess, we will still have to check the build logs keep an eye out for uncaught messages.


ntessore:
A number of documentation warnings should pop up. You can see them in the "checks" tab under the documentation build, and in the changed files.

PR #85, opened 2020-04-07
define models by parametrising random variables

ntessore:
This PR adds a mechanism to define models by parametrising random variables, such as those found in [`scipy.stats`](https://docs.scipy.org/doc/scipy/reference/stats.html). All of the r.v. functions such as `pdf`, `cdf`, etc. are made available. More importantly, the sampling with `rvs` is inherited from the r.v. and does not need to be reimplemented.

There is a suite of standard tests for distributions, so these do not need to be reimplemented either.

As an example, here is `skypy.galaxy.redshift.smail`, which is simply `scipy.stats.gengamma` with different parameters. [The examples section is auto-generated](https://683-221432358-gh.circle-artifacts.com/0/html/api/skypy.galaxy.redshift.smail.html#skypy.galaxy.redshift.smail).

```py
@examples(name='smail', args=(1.2, 1.5, 2.0))
def smail_par(z_median, alpha, beta):
    r'''Redshifts following the Smail et al (1994) model.

    ...

    '''

    a = (alpha+1)/beta
    c = beta
    loc = 0
    scale = z_median*scipy.special.gammainccinv(a, 0.5)**(-1/beta)

    return a, c, loc, scale


smail = parametrise(scipy.stats.gengamma, smail_par, name='smail')
```


ntessore:
I would be glad about suggestions, especially for the names of things and syntax.

ntessore:
I un-draft this, although it is not the finished article. But I think we can only see if and how this works in practice by using it, so I propose a WIP merge and we tune the mechanism a little as needed.

PR #88, opened 2020-04-09
organise GitHub actions

ntessore:
splits the current workflow up into three separate ones for tests, flake8, and documentation


ntessore:
> * You have an extra step `coverage xml -i` that wasn't in the CircleCI job - is this needed because your using `codecov-action` instead of just `codecov`?

Yep, the action takes an XML file and reports to codecov. The cookie cutter `--coverage` flag generates a HTML report, and `coverage xml -i` is what pytest-cov would call with the `--cov-report=xml` flag instead. I'm certain the default format could be changed from HTML to XML, but ...

> * We discussed before the Python 3.5 end of life is approaching pretty soon. Astropy also now require 3.6. I feel we may as well make the leap?

... I think that at some point in the near future we will want to upgrade the infrastructure to align with Astropy 4, and hand the building with different versions of Python etc. over to tox, where the default version for tests, and the coverage flags, are all handled.

PR #133, opened 2020-04-17
added Zenodo badge with DOI

itrharrison:
## Description
Added the Zenodo badge with the DOI of the latest release.

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -2,7 +2,7 @@
 SkyPy: A package for modelling the Universe
 ===========================================
 
-|Astropy Badge| |Test Status| |Coverage Status| |PyPI Status| |Documentation Status|
+|Zenodo Badge| |Astropy Badge| |Test Status| |Coverage Status| |PyPI Status| |Documentation Status|
 
 This package contains methods for modelling the Universe, galaxies and the Milky
 Way. Also included are methods for generating observed data.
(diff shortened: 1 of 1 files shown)


itrharrison:
We could link explicitly to the badge for the Concept DOI
```.. |Zenodo Badge| image:: https://zenodo.org/badge/doi/10.5281/zenodo.3755531.svg
   :target: https://doi.org/10.5281/zenodo.3755531
   :alt: DOI for all SkyPy Releases
```

Note that this is probably contrary to the spirit of the Zenodo guidelines:
https://help.zenodo.org/

> You should normally always use the DOI for the specific version of your record in citations. This is to ensure that other researchers can access the exact research artefact you used for reproducibility. By default, Zenodo uses the specific version to generate citations.
You can use the Concept DOI representing all versions in citations when it is desirable to cite an evolving research artifact, without being specific about the version.

But I do think my personal preference would be to use the Concept DOI.

rrjbca:
Do we count as an "evolving research artifact"

rrjbca:
Sorry, I actually inspected the badge more closely and you've used ':target: https://zenodo.org/badge/latestdoi/221432358' so I think that's what we want. By which I mean, it say 'latestdoi' in the target hyperlink

Issue #159, opened 2020-04-19
Improve Schechter function sampling


ntessore:
I am working on a number of fixes for the Schechter function sampling as currently implemented.

- Improved `gammaincc` that works for all negative indices. The current `upper_incomplete_gamma` quickly breaks, which is a problem when e.g. halo masses need to be sampled from the generalised Schechter function:
    ```py
    >>> from skypy.utils.special import upper_incomplete_gamma
    >>> upper_incomplete_gamma(-0.2, 1.0)  # 0.201155
    0.20115548418183707
    >>> upper_incomplete_gamma(-1.2, 0.5)  # 0.138937
    nan
    ```
    The algorithm I originally proposed is also rather wasteful in that it requires special functions in each iteration.

- Inverse function `gammainccinv` for negative indices. A direct inverse function for `gammaincc` that works for negative indices makes numerically integrating the CDF unnecessary. This allows us to get rid of the `resolution` argument, and have the option of sampling without upper limit, which is usually what's wanted, since the tail is exponentially suppressed anyway.

PR #164, opened 2020-04-20
improved gammaincc function

ntessore:
## Description

First part of #159: Replaces `upper_incomplete_gamma` with a `gammaincc(a, x)` function that returns the correct values for `a` real and `x` nonnegative. Tests for all edge cases and against values computed with Mathematica.

If speed becomes an issue it is likely possible to tune the loop for the recurrence.

before:
```py
>>> from skypy.utils.special import upper_incomplete_gamma
>>> upper_incomplete_gamma(-0.2, 1.0)  # 0.201155
0.20115548418183707
>>> upper_incomplete_gamma(-1.2, 0.5)  # 0.138937
nan
```
after:
```py
>>> from skypy.utils.special import gammaincc
>>> gammaincc(-0.2, 1.0)  # -0.034556
-0.03455597839676827
>>> gammaincc(-1.2, 1.0)  # 0.0286411
0.02864107572999873
```

## Checklist
- [x] Write unit tests


ntessore:
> * Perhaps save `GAMMAINCC_VALUES` in a date file? See `test_camb.py` for example using `get_pkg_data_filename`

I dislike having an unnecessary extra file which is only ever read in one place, but I do concede that's how Astropy does it. Will change.

> * Any reason for prefering `@raises(ValueError)` over `with pytest.raises(ValueError):`?

I find it easier if every test case checks only a single behaviour (after I spent an afternoon trying to fix the wrong numerical failure...), so having the function be the context where the exception is raised made sense to me.

philipp128:
> ## Description
> First part of #159: Replaces `upper_incomplete_gamma` with a `gammaincc(a, x)` function that returns the correct values for `a` real and `x` nonnegative. Tests for all edge cases and against values computed with Mathematica.
> 
> If speed becomes an issue it is likely possible to tune the loop for the recurrence.
> 
> before:
> 
> ```python
> >>> from skypy.utils.special import upper_incomplete_gamma
> >>> upper_incomplete_gamma(-0.2, 1.0)  # 0.201155
> 0.20115548418183707
> >>> upper_incomplete_gamma(-1.2, 0.5)  # 0.138937
> nan
> ```
> 
> after:
> 
> ```python
> >>> from skypy.utils.special import gammaincc
> >>> gammaincc(-0.2, 1.0)  # -0.034556
> -0.03455597839676827
> >>> gammaincc(-1.2, 1.0)  # 0.0286411
> 0.02864107572999873
> ```
> 
> ## Checklist
> * [x]  Follow the [Contributor Guidelines](https://github.com/skypyproject/skypy/blob/master/CONTRIBUTING.md)
> * [x]  Write unit tests
> * [x]  Write documentation strings
> * [x]  Assign someone from your working team to review this pull request
> * [x]  Assign someone from the infrastructure team to review this pull request

If I calculate the upper incomplete gamma function Gamma(-0.2, 1) I get 0.201155 and not -0.034556. I used WolframAlpha.

ntessore:
> Ah ok, sorry. I saw that you implemented the regularized upper inc. gamma function and 'corrected' this in the code by multiplying with the gamma function. If I am right you are taking the absolute values of gammainc because `scipy.special.gammaln` is calculating the logarithm of the absolute value of Gamma?

Yes, the _upper_ incomplete gamma function is positive, so the sign of the regularised gamma function is the sign of gamma, hence both absolute values.

skypy/galaxy/tests/test_luminosity.py
```diff
@@ -36,12 +36,10 @@ def calc_pdf(luminosity, z, alpha, a_m, b_m, absolute_magnitude_max=-16.0,
...
-        c = special.upper_incomplete_gamma(alpha + 1,
-                                           luminosity_min / lum_star)
-        d = special.upper_incomplete_gamma(alpha + 1,
-                                           luminosity_max / lum_star)
-        return 1. / lum_star * np.power(luminosity / lum_star, alpha) * np.exp(
-            - luminosity / lum_star) / (c - d)
+        c = special.gammaincc(alpha+1, luminosity_min/lum_star)
+        d = special.gammaincc(alpha+1, luminosity_max/lum_star)
+        return 1./lum_star*np.power(luminosity/lum_star, alpha)*np.exp(
+                -luminosity/lum_star)/(c - d)
```
philipp128, line 42: Here, (c-d) should also be multiplied with `Gamma(alpha+1)` using the regularised upper inc. gamma function, shouldn't, it?

ntessore: for the real PDF, yes, but the CDF is normalised so that is unnecessary

philipp128: True, but might be confusing in the future?

rrjbca, line 42: I suggest that we do normalise the pdf properly; somebody in the future might change the cdf integration method without realising the pdf is not normalised.
```suggestion
        from scipy.special import gamma
        c = special.gammainc(alpha+1, luminosity_min/lum_star) / gamma(alpha+1)
        d = special.gammainc(alpha+1, luminosity_max/lum_star) / gamma(alpha+1)
        return 1./lum_star*np.power(luminosity/lum_star, alpha)*np.exp(
                -luminosity/lum_star)/(c - d)
```

ntessore: That whole test is actually broken because the default upper and lower limits of the luminosity sampling do not coincide with the upper and lower limits given to the CDF. The explicit normalisation just papered that over. Will need a separate PR to fix.

philipp128: Oh ja I just checked it. Apparently, I forgot to change the test when we decided to use the units of L_*.

skypy/galaxy/tests/test_luminosity.py
```diff
@@ -1,6 +1,7 @@
 import numpy as np
 import scipy.stats
 import scipy.integrate
+import scipy.stats
```
rrjbca, line 4:
```suggestion
import scipy.special
```

Issue #168, opened 2020-04-20
align documentation with Astropy standard


ntessore:
Cause of #127. To use the same templates for generated stubs as Astropy, we need to use `automodapi` instead of `automodule`. This probably also means we should align the documentation more generally with Astropy.

- Module docstrings only contain a short summary.
- Package documentation gets written into the rST file.
- Reference/API section at the end of the package docs.

We should still be able to list the individual models above the API section, in the rST files. The autosummaries would not generate the toctree entries in that case.

PR #169, opened 2020-04-20
Astropyify galaxy docs

ntessore:
## Description

Part of #168. Overhauls the galaxy documentation to have a more meaningful list of galaxy properties, and the Astropy-like "Reference/API" section with `automodapi` generated documentation at the bottom. For this, the modules only contain a short summary sentence docstring, and exported symbols are declared in `__all__`. Actual documentation is written in the rST file.

This PR is meant as a test for this style of documentation. If accepted, the other existing packages will be converted as well.

## Checklist
- [x] Write unit tests


ntessore:
The issue with #168 is really that the Reference/API section is what calls `automodapi` which build the reference stubs using the automodapi (i.e. Astropy) templates, instead of the autosummary templates.

And while it may look duplicate at the moment, I thought the list at the top would be a ``list of all models'', which is a physical thing and will be surrounded by explanatory text eventually, while the Reference/API section is a resource for coding, and contains non-model entries as well (see `angular_size`).

rrjbca:
We definitely need to flesh out with explanatory text

ntessore:
I still think it is good to have the Astronomer's documentation and sectioning at the top, eventually with actual explanation, and the Programmer's reference manual at the bottom, where it is not nonsensical to separate objects by type.

If the majority is against keeping the Reference/API sections, the best way I see forward is then to try and make the templates we have in `docs/_templates/autosummary` (which inherit from the base auto**modapi** template) actually for auto**summary**. I am not sure what's going on there currently.

The issue for this is #168 so perhaps best to collect everything to do there.

Issue #177, opened 2020-04-29
Stellar-to-Halo Mass Relation


gfavole:
## Description
Build the stellar-to-halo mass relation from which we infer the conditional probability for central haloes with given mass of hosting a galaxy with a given stellar mass: P(Mstar|Mhalo)

## Inputs
 - halo mass
 - stellar mass

## Outputs
 - conditional probability P(Mstar|Mhalo)

## References
Birrer et al. 2018, arXiV:1401.3162

PR #181, opened 2020-05-04
Conditional probability function for the stellar-to-halo-matter relation

gfavole:
## Description
Conditional probability function to assign a galaxy stellar mass to a halo of a given mass built on the stellar-to-halo mass relation (SHMR) by Birrer et al. 2018, arXiV:1401.3162

This pull request addresses issue number #177 
## Inputs
- halo masses sampled from halo sampler
- stellar masses
## Outputs
- probabilities P(Mstar|Mhalo)

### DIFF
--- a/skypy/utils/random.py
+++ b/skypy/utils/random.py
@@ -11,7 +11,6 @@ Utility Functions
    :toctree: ../api/
 
    schechter
-
 """
 
 import numpy as np
@@ -59,7 +58,6 @@ def schechter(alpha, x_min, x_max, resolution=100, size=None):
     t_upper = np.interp(x_max, x, cdf)
     u = np.random.uniform(t_lower, t_upper, size=size)
     x_sample = np.interp(u, cdf, x)
-
     return x_sample
 
 
@@ -68,22 +66,19 @@ def _schechter_cdf(x, x_min, x_max, alpha):
     b = special.gammaincc(alpha + 1, x)
     c = special.gammaincc(alpha + 1, x_min)
     d = special.gammaincc(alpha + 1, x_max)
-
     return (a - b) / (c - d)
-    
-    
 
-	
-	
-def conditional_prob_shmr(x_min, x_max, size=None): 
+
+def conditional_prob_shmr(x_min, x_max, size=None):
     """ Stellar-to-halo- mass relation.
-       	Sample conditional probabilities P(Mstar|Mhalo) based on Birrer et al. 2018 
-       	stellar-to-halo-mass relation.
+    Sample conditional probabilities P(Mstar|Mhalo) based
+    on Birrer et al. 2018 stellar-to-halo-mass relation.
 
     Parameters
     ----------
     x_min, x_max : array_like
-        Lower and upper bounds for the random variable x representing the log halo mass.
+        Lower and upper bounds for the random variable x
+        representing the log halo mass.
     size: int, optional
         Output shape of samples. Default is None.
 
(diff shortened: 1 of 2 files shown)


skypy/utils/random.py
```diff
@@ -11,11 +11,12 @@
    :toctree: ../api/
 
    schechter
-
```
Lucia-Fonseca, line 14: The name of the additional functions should appear here, right after `schechter`  

```
   .. autosummary::
   :nosignatures:
    :toctree: ../api/

    schechter
    conditional_prob_shmr
```
 However, I think there are plans to change this structure. Hopefully, the infrastructure team can tell us.

skypy/utils/random.py
```diff
@@ -11,11 +11,12 @@
...
 """
 
 import numpy as np
 import skypy.utils.special as special
+from scipy import integrate
+import scipy.special as sp
```
Lucia-Fonseca, line 20: We need to specify the list of public objects of this module
```
__all__ = [
     'schechter',
     'conditional_prob_shrm',
 ]
```

skypy/utils/random.py
```diff
@@ -66,5 +66,39 @@ def _schechter_cdf(x, x_min, x_max, alpha):
...
     c = special.gammaincc(alpha + 1, x_min)
     d = special.gammaincc(alpha + 1, x_max)
-
     return (a - b) / (c - d)
+
+
+def conditional_prob_shmr(x_min, x_max, size=None):
+    """ Stellar-to-halo- mass relation.
+    Sample conditional probabilities P(Mstar|Mhalo) based
+    on Birrer et al. 2018 stellar-to-halo-mass relation.
```
Lucia-Fonseca, line 75: You could refer to the bibliography (it will provide a link directly to the reference in the documentation). Like this:
```
""" Stellar-to-halo- mass relation.
     Sample conditional probabilities P(Mstar|Mhalo) based
     on Birrer et al. 2018 stellar-to-halo-mass relation [1]_.
```
Mind the underscore after the number.

skypy/utils/random.py
```diff
@@ -66,5 +66,39 @@ def _schechter_cdf(x, x_min, x_max, alpha):
     b = special.gammaincc(alpha + 1, x)
     c = special.gammaincc(alpha + 1, x_min)
     d = special.gammaincc(alpha + 1, x_max)
-
     return (a - b) / (c - d)
+
+
+def conditional_prob_shmr(x_min, x_max, size=None):
```
Lucia-Fonseca, line 72: Sometimes, naming functions can be tricky. In SkyPy we tend to avoid abbreviations (although in your case it can be very cumbersome). I would suggest renaming the function: `conditional_probability_shmr`.
Q: is this function likely to be used within another context beyond SHMR? If so, maybe we could drop `shmr` from the name.
These are only suggestions.

skypy/utils/tests/test_random.py
```diff
@@ -34,3 +36,32 @@ def calc_cdf(x):
...
+    def calc_cdf(x):
+        x = np.linspace(x_min, x_max, 1000)
+        cdf = integrate.cumtrapz(sp.erf(x), x, initial=0)
+        cdf = cdf/cdf[-1]
+        return cdf
+    sample = random.conditional_prob_shmr(x_min=x_min, x_max=x_max, size=1000)
+
+    # Test the distribution of galaxy properties follows the right distribution
+    p_value = scipy.stats.kstest(sample, calc_cdf)[1]
+    assert p_value > 0.01
```
Lucia-Fonseca, line 67: Looks good!

rrjbca: This test is not working as intended. `kstest` takes samples and a cdf as arguments to compare. But you are passing it the same cdf twice since `conditional_prob_shmr` returns the cdf, not samples. The test appears to pass because the kstest calculates p_value = 1 but this is clearly incorrect.

rrjbca:
Speaking with @aamara he had proposed a model for mass quenching using an error function. See #183. Is this what you are implementing in this pull request?

rrjbca:
I think this was a duplicate of PR #191 which has merged so I am closing this PR now. If I've made a mistake then just let me know.

PR #197, opened 2020-06-23
Reshape of Power Spectra

### LINKED ISSUE #195: transpose power spectra
The power spectra are currently output which axes order `(nk, nz)`. This is the worse choice because it is always `P(k)` for a given `z` that is needed in computations, so the `P(k)` values for given `z` should always be contiguous in memory for the usual reasons. Which is also why every other power spectrum code returns `(nz, nk)`.

We had a vote and decided to change the order, which needs to happen before any code actually uses the power spectra.

Lucia-Fonseca:
## Description
This pull request reshapes the power spectra output from `(nk,nz)` to `(nz,nk)`. This closes #195.  

## Checklist
- [x] Write unit tests


Lucia-Fonseca:
@rrjbca could you help me understand why this is not passing all the checks? Thanks

skypy/power_spectrum/_halofit.py
```diff
@@ -98,18 +99,45 @@ def halofit(wavenumber, redshift, linear_power_spectrum,
...
-    >>> pvec = np.array([388.6725682632502, 0.21676249605280398])
     >>> cosmo = FlatLambdaCDM(H0=67.04, Om0=0.21479, Ob0=0.04895)
+
+    The Takahashi halofit power spectrum in :math:`Mpc^3` at redshift 0 and at
+    wavenumers :math:`k=0.0001/Mpc` and :math:`k=10.1/Mpc`:
+
+    >>> zvalue = 0.0
+    >>> kvec = np.array([1.00000000e-04, 1.01000000e+01])
+    >>> pvec = np.array([[388.6725682632502, 0.21676249605280398]])
     >>> halofit(kvec, zvalue, pvec, cosmo, _takahashi_parameters)
```
ntessore, line 116: Even though it was like that from before, I think this example would better use `halofit_takahashi` since the `_takahashi_parameters` are not exported from the module.

skypy/power_spectrum/_halofit.py
```diff
@@ -98,18 +99,45 @@ def halofit(wavenumber, redshift, linear_power_spectrum,
...
-    array([388.67064424,   0.72797614])
+    array([[388.67064424,   0.72797614]])
+
+    Create a plot for the Smith halofit model at redshift 0
+    and Planck15 cosmology:
+
+    >>> from skypy.power_spectrum import halofit
```
ntessore, line 122: This import needs to go to the top of the first example block to resolve the issue with `halofit` undefined in the docs.

skypy/power_spectrum/_halofit.py
```diff
@@ -98,18 +99,45 @@ def halofit(wavenumber, redshift, linear_power_spectrum,
...
-    >>> zvalue = 0.0
-    >>> pvec = np.array([388.6725682632502, 0.21676249605280398])
     >>> cosmo = FlatLambdaCDM(H0=67.04, Om0=0.21479, Ob0=0.04895)
+
+    The Takahashi halofit power spectrum in :math:`Mpc^3` at redshift 0 and at
+    wavenumers :math:`k=0.0001/Mpc` and :math:`k=10.1/Mpc`:
+
+    >>> zvalue = 0.0
+    >>> kvec = np.array([1.00000000e-04, 1.01000000e+01])
+    >>> pvec = np.array([[388.6725682632502, 0.21676249605280398]])
```
ntessore, line 115: I think the array shapes made more sense before because `np.shape(z)` is empty, so the power spectrum should have shape `(nk,)` instead of `(1, nk)`.

skypy/power_spectrum/_halofit.py
```diff
@@ -98,18 +99,45 @@ def halofit(wavenumber, redshift, linear_power_spectrum,
...
+    >>> plt.loglog(k,p.T, label='Linear')  # doctest: +SKIP
+    >>> plt.xlabel(r'k $(1/Mpc)$')  # doctest: +SKIP
+    >>> plt.ylabel(r'P $(Mpc^3)$')  # doctest: +SKIP
+    >>> plt.legend()  # doctest: +SKIP
+    >>> plt.show()  # doctest: +SKIP
+
     '''
 
     # Manage shapes of input arrays
-    return_shape = np.shape(linear_power_spectrum)
```
ntessore, line 112: I think it might be good to keep this because one would not expect the shape to change between input and output power spectrum, and the `linear_power_spectrum` is necessarily made 2d in the next step for the computations to work.

skypy/power_spectrum/_halofit.py
```diff
@@ -219,7 +247,7 @@ def log_sigma_squared(lnR):
     # Halofit non-linear power spectrum, Smith et al. 2003 equation C1
     pknl = 2 * np.pi * np.pi * (dq2 + dh2) / k3
 
-    return pknl.T.reshape(return_shape)
```
ntessore, line 222: see above

skypy/power_spectrum/tests/test_halofit.py
```diff
@@ -47,8 +51,8 @@ def test_halofit():
 
     # Test when redshift is a scalar
     z_scalar = z[0]
-    power_1d = linear_power[:, 0]
-    truth_scalar_redshift = truth_smith[:, 0]
+    power_1d = linear_power[0, :].reshape(1, len(k))
+    truth_scalar_redshift = truth_smith[0, :].reshape(1, len(k))
```
ntessore, line 55: Here the same, `z_scalar` has empty shape so the return array should be 1d.

skypy/power_spectrum/_halofit.py
Lucia-Fonseca: I agree. Let me look into this

Lucia-Fonseca:
`test_camb` seems to fail here when I run `tox -e test-scipy14`:
`>       assert pkz.shape == (len(redshift), len(wavenumber))
E       assert (2, 2) == (2, 200)
E         At index 1 diff: 2 != 200`
However, if I run the same code in a Jupyter notebook, there is no such error and `pkz.shape` is actually `(2,200)`.
Any idea why this happens?

Lucia-Fonseca:
Merge conflict resolved and all required checks passed.

skypy/power_spectrum/_camb.py
```diff
@@ -1,3 +1,9 @@
+"""CAMB module.
+
+This module facilitates the CAMB computation of the linear
+matter power spectrum.
+"""
+
```
ntessore, line 6: This is just a private implementation file and not a submodule.

skypy/power_spectrum/_camb.py
```diff
@@ -8,13 +14,14 @@
...
-    two dimensional grid of wavenumber and redshift
+    r'''CAMB linear matter power spectrum.
+    Return the CAMB computation of the linear matter power spectrum, on a
+    two dimensional grid of wavenumber and , described in [1]_.
 
     Parameters
     ----------
     wavenumber : (nk,) array_like
-        Array of wavenumbers in units of [Mpc^-1] at which to
+        Array of wavenumbers in units of :math:`[Mpc^-1]` at which to
```
ntessore, line 24: Does not render properly: image

skypy/power_spectrum/_camb.py
```diff
@@ -31,14 +38,19 @@ def camb(wavenumber, redshift, cosmology, A_s, n_s):
 
     Returns
     -------
-    power_spectrum : (nk, nz) array_like
-        Array of values for the linear matter power spectrum in  [Mpc^3]
+    power_spectrum : (nz, nk) array_like
+        Array of values for the linear matter power spectrum in :math:`[Mpc^3]`
```
ntessore, line 42: See above.

skypy/power_spectrum/_camb.py
```diff
@@ -31,14 +38,19 @@ def camb(wavenumber, redshift, cosmology, A_s, n_s):
...
         evaluated at the input wavenumbers for the given primordial power
         spectrum parameters, cosmology. For nz redshifts and nk wavenumbers
-        the returned array will have shape (nk, nz).
+        the returned array will have shape (nz, nk).
 
     Examples
     --------
+
+    This will return the linear matter power spectrum in :math:`Mpc^3`
+    at several values of redshift and wavenumers in :math:`1/Mpc`
```
ntessore, line 51: See above.

skypy/power_spectrum/_halofit.py
```diff
@@ -1,3 +1,9 @@
+"""HALOFIT module.
+
+This module facilitates the computation of the halofit matter power spectrum.
+"""
+
+
```
ntessore, line 6: See above.

skypy/power_spectrum/_halofit.py
```diff
@@ -71,11 +77,12 @@ def halofit(wavenumber, redshift, linear_power_spectrum,
     Parameters
     ----------
     k : (nk,) array_like
-        Input wavenumbers in units of [Mpc^-1].
+        Input wavenumbers in units of :math:`[Mpc^-1]`.
```
ntessore, line 80: See above.

skypy/power_spectrum/_halofit.py
```diff
@@ -71,11 +77,12 @@ def halofit(wavenumber, redshift, linear_power_spectrum,
...
     z : (nz,) array_like
         Input redshifts
-    P : (nk, nz) array_like
-        Linear power spectrum for given wavenumbers and redshifts [Mpc^3].
+    P : (nz, nk) array_like
+        Linear power spectrum for given wavenumbers and
+        redshifts :math:`[Mpc^3]`.
```
ntessore, line 85: See above.

skypy/power_spectrum/_halofit.py
```diff
@@ -84,8 +91,8 @@ def halofit(wavenumber, redshift, linear_power_spectrum,
 
     Returns
     -------
-    pknl : (nk, nz) array_like
-           Non-linear halo power spectrum in units of [Mpc^3].
+    pknl : (nz, nk) array_like
+           Non-linear halo power spectrum in units of :math:`[Mpc^3]`.
```
ntessore, line 95: See above.

skypy/power_spectrum/_halofit.py
```diff
@@ -98,21 +105,32 @@ def halofit(wavenumber, redshift, linear_power_spectrum,
...
     '''
 
     # Manage shapes of input arrays
     return_shape = np.shape(linear_power_spectrum)
     redshift = np.atleast_1d(redshift)
+
     if np.ndim(linear_power_spectrum) == 1:
-        linear_power_spectrum = linear_power_spectrum[:, np.newaxis]
+        length_pk = len(linear_power_spectrum)
+        linear_power_spectrum = linear_power_spectrum.reshape(1, length_pk)
```
ntessore, line 133:
```suggestion
        linear_power_spectrum = linear_power_spectrum.reshape(1, -1)
```

Lucia-Fonseca: Why this change?

ntessore: it seems rather pointless to get the length explicitly just to add a dimension -- alternatives are having a `None` index, or using `expand_dims`, but I think the implicit reshape explains best what happens

Lucia-Fonseca: makes sense! Thanks

skypy/power_spectrum/_camb.py
```diff
@@ -98,4 +110,9 @@ def camb(wavenumber, redshift, cosmology, A_s, n_s):
                                                    maxkh=np.max(k_h.value),
                                                    npoints=len(k_h.value))
 
-    return pzk[redshift_order[::-1]].T
+    if len(redshift_order) > 1:
+        power_output = pzk[redshift_order[::-1]]
+    else:
+        power_output = pzk[redshift_order[::-1]].reshape(return_shape)
+
+    return power_output
```
ntessore, line 118: Why is this not just `return pzk[redshift_order[::-1]]` i.e. as before without the transposition?

Lucia-Fonseca: let me check

Lucia-Fonseca: If I just do 
`return pzk[redshift_order[::-1]]`
the shape of the power spectrum is `(1, 100)` for 100 values of `k` and NOT `(100,)`.
I am sure there is a more clever way to do it right. Do you have any suggestion?

ntessore: You can `np.squeeze` the result, although then you get shape `(100,)` even if redshift is an array with one entry. If you want to get it exactly right, you can do `.reshape(*np.shape(redshifts), *np.shape(wavenumber))` which is ugly but should correctly differentiate scalars and one-entry array input.

Lucia-Fonseca: Hmm that doesn't seem to work either.

skypy/power_spectrum/_camb.py
rrjbca:
```suggestion
        Array of wavenumbers in units of :math:`[{\rm Mpc}^{-1}]` at which to
```
Perhaps try this? That is what I would do in LaTeX

rrjbca:
Things like changing how `Mpc` renders, adding references and similar documentation changes have nothing to do with this PR. I am going to revert the Mpc changes that are not working and open a new issue. I will merge the other changes this time but please try to keep your pull requests focused.

ntessore (approved):

PR #203, opened 2020-06-26
test matrix

ntessore:
## Description

Tests across a range of python, astropy, numpy, and scipy versions.

## Checklist
- [x] Write unit tests

### DIFF
--- a/.github/workflows/tests.yaml
+++ b/.github/workflows/tests.yaml
@@ -9,7 +9,6 @@ on:
 jobs:
   test:
     name: ${{ matrix.name }}
-    continue-on-error: ${{ !matrix.required }}
     runs-on: ${{ matrix.os }}
     strategy:
       fail-fast: false
@@ -17,7 +16,6 @@ jobs:
         include:
 
           - name: 🐣 latest supported versions
-            required: true
             os: ubuntu-latest
             python: 3.8
             toxenv: py38-test-all-latest-cov
@@ -25,25 +23,21 @@ jobs:
             coverage: true
 
           - name: 🏚 oldest supported versions
-            required: false
             os: ubuntu-latest
             python: 3.6
             toxenv: py36-test-oldest
 
           - name: 🧪 latest experimental versions
-            required: false
             os: ubuntu-latest
             python: 3.8
             toxenv: py38-test
 
           - name: 🍏 macOS latest
-            required: false
             os: macos-latest
             python: 3.8
             toxenv: py38-test-all-latest
 
           - name: 💻 Windows latest
-            required: false
             os: windows-latest
             python: 3.8
             toxenv: py38-test-all-latest
--- a/.github/workflows/tests.yaml
+++ b/.github/workflows/tests.yaml
@@ -27,22 +27,12 @@ jobs:
             python: 3.6
             toxenv: py36-test-oldest
 
-          - name: latest release versions
-            os: ubuntu-latest
-            python: 3.8
-            toxenv: py38-test
-
-          - name: latest experimental versions
-            os: ubuntu-latest
-            python: 3.8
-            toxenv: py38-test-dev
-
-          - name: macOS latest
+          - name: macOS latest supported
             os: macos-latest
             python: 3.8
             toxenv: py38-test-latest
 
-          - name: Windows latest
+          - name: Windows latest supported
             os: windows-latest
             python: 3.8
             toxenv: py38-test-latest
--- a/.github/workflows/tests.yaml
+++ b/.github/workflows/tests.yaml
@@ -9,43 +9,43 @@ on:
 jobs:
   test:
     name: ${{ matrix.name }}
+    continue-on-error: ${{ !matrix.required }}
     runs-on: ${{ matrix.os }}
-    continue-on-error: ${{ matrix.experimental }}
     strategy:
       fail-fast: false
       matrix:
         include:
 
           - name: latest supported versions
+            required: true
             os: ubuntu-latest
             python: 3.8
-            experimental: false
-            coverage: true
             toxenv: py38-test-all-latest-cov
             toxposargs: --cov-report xml
+            coverage: true
 
           - name: oldest supported versions
+            required: true
             os: ubuntu-latest
             python: 3.6
-            experimental: false
             toxenv: py36-test-oldest
 
           - name: latest experimental versions
+            required: false
             os: ubuntu-latest
             python: 3.8
-            experimental: true
             toxenv: py38-test
 
           - name: macOS latest
+            required: false
             os: macos-latest
             python: 3.8
-            experimental: true
             toxenv: py38-test-all-latest
 
           - name: Windows latest
+            required: false
             os: windows-latest
             python: 3.8
-            experimental: true
             toxenv: py38-test-all-latest
 
     steps:
--- a/.github/workflows/tests.yaml
+++ b/.github/workflows/tests.yaml
@@ -11,10 +11,10 @@ jobs:
     runs-on: ubuntu-latest
     strategy:
       matrix:
-        python: [3.6, 3.7, 3.8]
+        python: [3.6, 3.8]
         astropy: [astropy30, astropy40]
-        numpy: [numpy116, numpy117, numpy118]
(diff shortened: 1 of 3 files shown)


ntessore:
The `oldest supported versions` are currently not marked as required because we need to fix a number of issues already present in the code for them to pass. Once those are sorted out, we can mark the test as required.

.github/workflows/tests.yaml
```diff
@@ -8,23 +8,57 @@ on:
...
+
+          - name: oldest supported versions
+            os: ubuntu-latest
+            python: 3.6
+            toxenv: py36-test-oldest
+
+          - name: latest experimental versions
+            os: ubuntu-latest
+            python: 3.8
+            toxenv: py38-test
```
rrjbca, line 33:
```suggestion
            toxenv: py38-test-dev
```

.github/workflows/tests.yaml
```diff
@@ -8,23 +8,57 @@ on:
...
+
+          - name: macOS latest
+            os: macos-latest
+            python: 3.8
+            toxenv: py38-test-all-latest
```
rrjbca, line 38:
```suggestion
            toxenv: py38-test-latest
```

.github/workflows/tests.yaml
```diff
@@ -8,23 +8,57 @@ on:
...
+
+          - name: Windows latest
+            os: windows-latest
+            python: 3.8
+            toxenv: py38-test-all-latest
```
rrjbca, line 43:
```suggestion
            toxenv: py38-test-latest
```

tox.ini
```diff
@@ -1,9 +1,9 @@
 [tox]
 envlist =
-    py{36,37,38}-test{,-alldeps,-devdeps}{,-cov}
-    py{36,37,38}-test-numpy{116,117,118}
+    py{36,37,38}-test{,-all}{,-dev,-latest,-oldest}{,-cov}
+    py{36,37,38}-test-numpy{116,117,118,119}
     py{36,37,38}-test-scipy{12,13,14,15}
```
rrjbca, line 5:
```suggestion
```

tox.ini
```diff
@@ -1,9 +1,9 @@
...
-    py{36,37,38}-test-astropy{30,40,lts}
+    py{36,37,38}-test-astropy{32,40,lts}
```
rrjbca, line 6:
```suggestion
```

tox.ini
```diff
@@ -52,23 +54,33 @@ deps =
...
 
-    devdeps: git+https://github.com/numpy/numpy.git#egg=numpy
-    devdeps: git+https://github.com/astropy/astropy.git#egg=astropy
+    dev: git+https://github.com/numpy/numpy.git#egg=numpy
+    dev: git+https://github.com/scipy/scipy.git#egg=scipy
+    dev: git+https://github.com/astropy/astropy.git#egg=astropy
+
+    latest: astropy==4.0.*
+    latest: numpy==1.18.*
+    latest: scipy==1.4.*
```
rrjbca, line 74:
```suggestion
    latest: scipy==1.5.*
```

tox.ini
```diff
@@ -31,18 +31,20 @@ changedir = .tmp/{envname}
...
+    oldest: using the oldest supported version of key dependencies
     cov: and test coverage
     numpy116: with numpy 1.16.*
     numpy117: with numpy 1.17.*
     numpy118: with numpy 1.18.*
+    numpy119: with numpy 1.19.*
     scipy12: with scipy 1.2.*
     scipy13: with scipy 1.3.*
     scipy14: with scipy 1.4.*
     scipy15: with scipy 1.5.*
```
rrjbca, line 46:
```suggestion
```

tox.ini
```diff
@@ -31,18 +31,20 @@ changedir = .tmp/{envname}
...
-    astropy30: with astropy 3.0.*
+    astropy32: with astropy 3.2.*
     astropy40: with astropy 4.0.*
     astropylts: with the latest astropy LTS
```
rrjbca, line 49:
```suggestion
```

tox.ini
```diff
@@ -52,23 +54,33 @@ deps =
     numpy116: numpy==1.16.*
     numpy117: numpy==1.17.*
     numpy118: numpy==1.18.*
+    numpy119: numpy==1.19.*
 
     scipy12: scipy==1.2.*
     scipy13: scipy==1.3.*
     scipy14: scipy==1.4.*
     scipy15: scipy==1.5.*
```
rrjbca, line 63:
```suggestion
```

tox.ini
```diff
@@ -52,23 +54,33 @@ deps =
...
 
-    astropy30: astropy==3.0.*
+    astropy32: astropy==3.2.*
     astropy40: astropy==4.0.*
     astropylts: astropy==4.0.*
```
rrjbca, line 66:
```suggestion
```

.github/workflows/tests.yaml
ntessore: I have tried that, but it needs dependencies for building the modules. I don't think we need to support the dev versions necessarily, and I think it's ok if, most of the time, 'latest supported' and 'latest experimental' are the same versions.

tox.ini
ntessore: These specific pinned versions might be helpful for testing where exactly something breaks, but I also see why less might be more. Either way ok with me, but they do serve a purpose.

tox.ini
```diff
@@ -52,23 +54,33 @@ deps =
...
 
-    devdeps: git+https://github.com/numpy/numpy.git#egg=numpy
-    devdeps: git+https://github.com/astropy/astropy.git#egg=astropy
+    dev: git+https://github.com/numpy/numpy.git#egg=numpy
+    dev: git+https://github.com/scipy/scipy.git#egg=scipy
+    dev: git+https://github.com/astropy/astropy.git#egg=astropy
+
+    latest: astropy==4.0.*
+    latest: numpy==1.18.*
```
ntessore, line 73:
```suggestion
    latest: numpy==1.19.*
```

ntessore:
> * Can we add `[options.extras_require] all = camb` in `setup.cfg` and try to get that to work?

@itrharrison do you know what exactly is required to get a pip-installed camb working?

> * There is a `linkcheck` environment from the astropy template but I don't see it used for anything?

Good catch, that should be added to the docs workflow. I wanted to update that and code style check with a better name anyway, but separately.

tox.ini
rrjbca: OK sure lets keep them for now then

rrjbca:
In the astropy `tox.ini` there is an additional line relating to `linkcheck` that is missing from this PR:
```
deps =

    # Linkcheck is broken with cryptic error with Sphinx 3.1.0
    # https://github.com/astropy/astropy/issues/10489
    linkcheck: sphinx<3.1
```

tox.ini
rrjbca, line 6:
```suggestion
    py{36,37,38}-test-astropy{32,40}
```
Maybe only need astropy40 since astropylts resolves to the same?

ntessore: That all comes from the package template, where I would guess that astropy will let `40` evolve and keep `lts` pinned to 4.0. Anyway, if we adopt the oldest-latest testing strategy, we need pinned versions between those, and don't have to single out `lts`, and it can go.

rrjbca: Ah OK then I don't mind removing it or sticking to the template

.github/workflows/tests.yaml
rrjbca: Numpy and scipy build 'nightly' test wheels: https://anaconda.org/scipy-wheels-nightly which is what astropy use in their `devdeps` testing environment

rrjbca: I'd much rather try and get that to work so there is a meaningful difference between `latest` and `dev`

ntessore: I agree on getting `dev` to work, I'm not convinced about testing against unstable nightlies. I think the latest unsupported tests are helpful when new versions of numpy, scipy, astropy come out and we need to see where we stand, as happened now with scipy 1.5.

PR #211, opened 2020-07-09
Sampling of the galaxy stellar masses

philipp128:
## Description
This module samples the galaxy stellar masses from a Schechter function according to the model of [1].
## Checklist
- [ ] Write unit tests

## References
[1] Peng et al. (2010)

### DIFF
--- a/skypy/galaxy/mass.py
+++ b/skypy/galaxy/mass.py
@@ -1,0 +1,97 @@
+r"""Models of galaxy luminosities.
+
+"""
+
+import numpy as np
+
+from skypy.utils.random import schechter
+
+
+__all__ = [
+    'peng_masses',
+]
+
+
+def peng_masses(redshift, alpha, m_star, size=None,
+                x_min=0.0002138, x_max=213.8, resolution=100):
+    r"""Model of Peng et al (2010)
+
+    Stellar masses following the Schechter mass function following the
+    Peng et al. [1]_ model.
+
+    Parameters
+    ----------
+    redshift : (nz,) array-like
+        The redshift values at which to sample luminosities.
+    alpha : float or int
+        The alpha parameter in the Schechter stellar mass function.
+    m_star : float or int
+        Characteristic stellar mass M_*.
+    size: int, optional
+         Output shape of stellar mass samples. If size is None and redshift
+         is a scalar, a single sample is returned. If size is None and
+         redshift is an array, an array of samples is returned with the same
+         shape as redshift.
+    x_min, x_max : float or int, optional
+        Lower and upper stella mass bounds in units of M_*.
+    resolution : int, optional
+        Resolution of the inverse transform sampling spline. Default is 100.
+
+    Returns
+    -------
+    stellar mass : array_like
+        Drawn stellar masses from the Schechter stellar mass function in units
+        of the solar mass
+
+    Notes
+    -----
+    The Schechter luminosity function is given as
+
+    .. math::
+        \Phi(M, z) = \frac{\Phi_\star(z)}{M_\star}
+            \left(\frac{M}{M_\star}\right)^\alpha
+            \exp\left(-\frac{M}{M_\star}\right) \;.
+
+   The redshift dependence of :math:`\Phi_\star(z)` is given by
+
+    .. math::
+
+        \Phi_\star(z) = a_\phi z + b_phi \;.
+
+    Now we have to rescale the Schechter function by the comoving element and
+    get
+
+    .. math::
+
+        \phi(M,z) = \frac{d_H d_M^2}{E(z)}  \Phi(M,z)\;.
+
+    References
+    ----------
+    .. [1] Peng Y., Lilly S. J. et al., 2010, The Astrophysical Journal,
+    Issue 1, Volume 721, pages 193-221
+
+    Examples
+    --------
+    >>> import skypy.galaxy.mass as mass
+
+    Sample 100 stellar masses values at redshift z = 1.0 with alpha = -1.4, and
+    m_star = 10**10.67.
+
+    >>> masses = mass.peng_masses(1.0, -1.4, 10**10.67, size=100)
+
+    Sample a luminosity value for every redshift in an array z with
+    alpha = -1.3 amd m_star = 10**10.67.
+
+    >>> z = np.linspace(0,2, 100)
+    >>> masses = mass.peng_masses(z, -1.4, m_star = 10**10.67)
+
+
+
+    """
+
+    if size is None and np.shape(redshift):
+        size = np.shape(redshift)
+
+    x_sample = schechter(alpha, x_min, x_max, resolution=resolution, size=size)
+
+    return m_star * x_sample
--- a/skypy/galaxy/stellar_mass.py
+++ b/skypy/galaxy/stellar_mass.py
@@ -1,4 +1,4 @@
-r"""Models of galaxy mass (total and stellar).
+r"""Models of galaxy stellar mass.
 
 """
 
@@ -8,13 +8,13 @@ from skypy.utils.random import schechter
 
 
 __all__ = [
-    'stellar_mass_function',
+    'schechter_mass',
 ]
 
 
-def stellar_mass_function(alpha, m_star, size=None,
-                          x_min=0.0002138, x_max=213.8, resolution=100):
-    r""" Stellar masses following the Schechter mass function.
+def schechter_mass(alpha, m_star, size=None,
+                   x_min=0.0002138, x_max=213.8, resolution=100):
+    r""" Stellar masses following the Schechter mass function [1]_.
 
     Parameters
     ----------
@@ -62,6 +62,9 @@ def stellar_mass_function(alpha, m_star, size=None,
 
     References
     ----------
+    .. [1] Mo, H., Van den Bosch, F., & White, S. (2010). Galaxy Formation and
+    Evolution. Cambridge: Cambridge University Press.
+    doi:10.1017/CBO9780511807244
 
     Examples
     --------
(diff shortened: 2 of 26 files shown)


skypy/galaxy/tests/test_mass.py
```diff
@@ -0,0 +1,56 @@
...
+                              x_min=1e-10, x_max=q_max,
+                              resolution=1000)
+    d, p_value = scipy.stats.kstest(sample, 'truncexpon', args=(q_max,))
+    assert p_value >= 0.01
+
+
+def test_peng_masses():
+    # Test that error is returned if redshift input is an array but size !=
+    # None and size != redshift,size
+    with pytest.raises(ValueError):
```
philipp128, line 24: This test is failing and I don't understand why. It's basically the same code as in the case of sampling the luminosities.

skypy/galaxy/mass.py
```diff
@@ -0,0 +1,97 @@
...
+from skypy.utils.random import schechter
+
+
+__all__ = [
+    'peng_masses',
+]
+
+
+def peng_masses(redshift, alpha, m_star, size=None,
+                x_min=0.0002138, x_max=213.8, resolution=100):
```
rrjbca, line 16:
```suggestion
def peng_masses(alpha, m_star, size=None,
                x_min=0.0002138, x_max=213.8, resolution=100):
```

skypy/galaxy/mass.py
```diff
@@ -0,0 +1,97 @@
...
+    r"""Model of Peng et al (2010)
+
+    Stellar masses following the Schechter mass function following the
+    Peng et al. [1]_ model.
+
+    Parameters
+    ----------
+    redshift : (nz,) array-like
+        The redshift values at which to sample luminosities.
```
rrjbca, line 25:
```suggestion
```

skypy/galaxy/mass.py
```diff
@@ -0,0 +1,97 @@
...
+    alpha : float or int
+        The alpha parameter in the Schechter stellar mass function.
+    m_star : float or int
```
rrjbca, line 28:
```suggestion
    m_star : array-like
```

skypy/galaxy/mass.py
```diff
@@ -0,0 +1,97 @@
...
+        Characteristic stellar mass M_*.
+    size: int, optional
+         Output shape of stellar mass samples. If size is None and redshift
+         is a scalar, a single sample is returned. If size is None and
+         redshift is an array, an array of samples is returned with the same
+         shape as redshift.
```
rrjbca, line 34:
```suggestion
         Output shape of stellar mass samples. If size is None and m_star
         is a scalar, a single sample is returned. If size is None and
         m_star is an array, an array of samples is returned with the same
         shape as m_star.
```

skypy/galaxy/mass.py
```diff
@@ -0,0 +1,97 @@
...
+
+    >>> z = np.linspace(0,2, 100)
+    >>> masses = mass.peng_masses(z, -1.4, m_star = 10**10.67)
+
+
+
+    """
+
+    if size is None and np.shape(redshift):
+        size = np.shape(redshift)
```
rrjbca, line 93:
```suggestion
    if size is None and np.shape(m_star):
        size = np.shape(m_star)
```

skypy/galaxy/tests/test_mass.py
```diff
@@ -0,0 +1,56 @@
...
+        mass.peng_masses(np.array([1, 2]), -1.4, 10**10.67, size=3)
```
rrjbca, line 25:
```suggestion
    # Test that error is returned if m_star input is an array but size !=
    # None and size != m_star.size
    with pytest.raises(ValueError):
        mass.peng_masses(-1.4, np.array([1e10, 2e10]), size=3)
```

philipp128:
> There is no redshift dependence in this model so it doesn't make sense for this to be a function of redshift and the failing test is meaningless. However, we should generalise the function to accept m_star as an array. Then it would make sense for the function to raise an exception if size != m_star.size. I've given suggestions for how to make this change but you will need to update the rest of your examples and tests as well to make them consistent.

True. I am not sure whether it makes sense to make `m_star` array-like but not `alpha`?

rrjbca:
This suggestion is equivalent to what we did for the luminosity function where L* was an array and alpha was constant. The only difference in that case was that instead of passing L* to the function directly we passed an array of redshifts and calculated L* at each redshift inside the function.

skypy/galaxy/mass.py
```diff
@@ -0,0 +1,97 @@
+r"""Models of galaxy luminosities.
```
ntessore, line 1:
```suggestion
r"""Models of galaxy mass (total and stellar).
```

If there will not be total galaxy mass in here, then `mass.py` is the wrong name for this file.

skypy/galaxy/mass.py
```diff
@@ -0,0 +1,97 @@
...
+
+from skypy.utils.random import schechter
+
+
+__all__ = [
+    'peng_masses',
+]
+
+
+def peng_masses(redshift, alpha, m_star, size=None,
```
ntessore, line 15:
```suggestion
def stellar_mass_function(redshift, alpha, m_star, size=None,
```

This is the common Schechter stellar mass function as far as I can tell.

skypy/galaxy/mass.py
```diff
@@ -0,0 +1,97 @@
...
+                x_min=0.0002138, x_max=213.8, resolution=100):
+    r"""Model of Peng et al (2010)
+
+    Stellar masses following the Schechter mass function following the
+    Peng et al. [1]_ model.
```
ntessore, line 20: Same, what's makes this the Peng et al. (2010) model?

skypy/galaxy/mass.py
```diff
@@ -0,0 +1,97 @@
...
+
+    Returns
+    -------
+    stellar mass : array_like
+        Drawn stellar masses from the Schechter stellar mass function in units
+        of the solar mass
+
+    Notes
+    -----
+    The Schechter luminosity function is given as
```
ntessore, line 48:
```suggestion
    The stellar mass function is given by a Schechter function as
```

skypy/galaxy/mass.py
philipp128: Ah ok. I just wrote Peng, cause that's where I saw it the first time.

skypy/galaxy/mass.py
philipp128: Not yet. Maybe it comes in the future? I can change the name tho and if there will be total masses at some point, we can change it back.

rrjbca: The dark matter masses are handled in the halo module. So this should probably be stellar_mass.py or similar

ntessore: If the subpackage is `stellar_mass` then the function should probably be `schechter`. The other SMF I could think of is `broken_power_law`, so that `galaxy.stellar_mass.{schechter,broken_power_law}` makes more sense than `galaxy.stellar_mass.stellar_mass_function`.

philipp128: Ok this is true. I will change it then to this in the next commit.

philipp128: `schechter` already exists in `utils.random` and I am importing this. This gives an error with flake8 but also it's not really nice to have the same name for e more than one function in SkyPy, right? What do you think about `schechter_mass`?

Lucia-Fonseca: we might not need the Schechter function in `utils.random` after I push all the code in PR#186. We might need to discuss this offline at some point

rrjbca: I'm pretty sure we will want to keep `utils.random.schechter`. You could change how you import it to prevent the collision:
```
from skypy.utils import random
def schechter(...):
    x_sample = random.schechter(...)

Comment (philipp128) [reply skypy/galaxy/mass.py]: But is it nice to have more functions called 'schechter'? Also, it might be a bit misleading to call it just 'schechter', cause it's not a schechter function. Although it's in `stellar_mass.py` I don't really like it just calling schechter tbh.

**Comment by rrjbca** [comment]
We also need a test where `size = None` and m_star is an array to improve the test coverage. E.g. something that checks `m_star.shape == schechter_mass(..., m_star, ...).shape`

File: skypy/galaxy/mass.py (hunk 8 above)
Comment (ntessore) [reply skypy/galaxy/mass.py]: You can call it `schechter_smf`, and we can at some point rename the other function to `schechter_lf`. I think that both functions would ideally reside in the same module eventually, since they both deal with the luminous matter of a galaxy (eg. `skypy.galaxy.stars`) and there's little point in fracturing the module hierarchy more than that.

File: skypy/galaxy/stellar_mass.py [hunk 12]
```diff
@@ -0,0 +1,73 @@
...
+        of the solar mass
+
+    Notes
+    -----
+     The stellar mass function is given by a Schechter function as
+
+    .. math::
+        \Phi(M, z) = \frac{\Phi_\star(z)}{M_\star}
+            \left(\frac{M}{M_\star}\right)^\alpha
+            \exp\left(-\frac{M}{M_\star}\right) \;.
```
ntessore, line 47: Not really relevant here since that is the number density, but the sampling does not draw galaxies from a Schechter function, it assigns masses to already drawn galaxies. This needs more explanation so that users understand what this function is doing and how it interacts with the actual sampling of galaxies, which currently happens in the redshift module.

skypy/galaxy/stellar_mass.py
```diff
@@ -0,0 +1,73 @@
...
+
+from skypy.utils.random import schechter
+
+
+__all__ = [
+    'schechter_mass',
+]
+
+
+def schechter_mass(alpha, m_star, x_min, x_max, resolution=100, size=None):
```
ntessore, line 15: Confusing name since this is only stellar mass.

ntessore:
Looks okay now, but the new module is not included in the documentation yet.

philipp128:
> Looks okay now, but the new module is not included in the documentation yet.

How do you do this again? Which file is it?

PR #233, opened 2020-07-15
Remove related_identifiers from zenodo config

rrjbca:
## Description
The `related_identifiers` field in the zenodo config file is probably the cause of an error when creating a new version. See https://github.com/zenodo/zenodo/issues/1667 for a similar case. Removing that field from our config file.

## Checklist

### DIFF
--- a/.zenodo.json
+++ b/.zenodo.json
@@ -123,12 +123,5 @@
             "type": "ProjectMember"
         }
     ],
-    "access_right": "open",
-    "related_identifiers": [
-        {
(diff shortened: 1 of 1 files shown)


rrjbca:
I was able to create a release on the zenodo test server for a dummy repository using a similar config file so this fix should be successful:
- https://github.com/rrjbca/GitHubApps/blob/zenodo/.zenodo.json
- https://sandbox.zenodo.org/record/647647

PR #239, opened 2020-07-20
config and overwrite command line arguments for driver script

### LINKED ISSUE #232: Rerun yaml file requires to delete the output files
**Describe the bug**
Re-running `.yaml ` files returns an error with the message that the output files already exist. 

**To Reproduce**
```
skypy -c examples/herbel_galaxies.yaml -f fits
```
```
skypy -c examples/herbel_galaxies.yaml -f fits
```

**Expected behavior**
It should run without problems and updating the output files.

rrjbca:
## Description
- Make `config` a positional argument instead of a required argument
- Add `-o/--overwrite` command line flag to `skypy` script to overwrite any existing files with the same names as the output files without warning. Resolves #232 

## Checklist
- [x] Write unit tests

### DIFF
--- a/skypy/pipeline/tests/test_driver.py
+++ b/skypy/pipeline/tests/test_driver.py
@@ -21,13 +21,14 @@ def test_driver():
 
     # Generate a simple two column table with a dependency. Also write the
     # table to a fits file and check it's contents.
+    size = 100
     config = {'tables': {
                 'test_table': {
                   'column1': {
                     'module': 'numpy.random',
                     'function': 'uniform',
                     'kwargs': {
-                      'size': 100}},
+                      'size': size}},
                   'column2': {
                     'module': 'numpy.random',
                     'function': 'uniform',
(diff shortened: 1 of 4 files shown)


skypy/pipeline/tests/test_driver.py
```diff
@@ -41,6 +41,17 @@ def test_driver():
...
+    # Check for failure if output files already exist and overwrite is False
+    driver = SkyPyDriver()
+    with pytest.raises(OSError):
+        driver.execute(config, file_format='fits', overwrite=False)
+
+    # Check that the existing output files are modified if overwrite is True
+    timestamp = os.path.getmtime("test_table.fits")
+    driver = SkyPyDriver()
+    driver.execute(config, file_format='fits', overwrite=True)
+    assert(os.path.getmtime("test_table.fits") > timestamp)
```
ntessore, line 53: This might not have the time resolution if `getmtime` return seconds since the epoch? Waiting explicitly might help.

ntessore: Or, easier, make timestamp now and check that mtime >= now

rrjbca: I get sub-second precision, but that could be implementation/OS dependent

ntessore: Yeah the python docs only say 

> os.path.getmtime(path)
Return the time of last modification of path. The return value is a number giving the number of seconds since the epoch (see the time module). Raise os.error if the file does not exist or is inaccessible.

so I am not sure what we can infer from that.

rrjbca: Instead I now change `size` in the config file and check that the table in the file has the correct length i.e. it has been overwritten with new data

PR #244, opened 2020-08-05
refactor Schechter luminosity function sampling

ntessore:
## Description

This is an attempt to refactor the Schechter luminosity function sampling to use arbitrary models for `phi_star` and `M_star`.

This PR is based on #271 (but does not require it), so it should be merged after.

## Checklist
- [x] Write unit tests

### DIFF
--- a/examples/abundance_matching.yml
+++ b/examples/abundance_matching.yml
@@ -2,6 +2,7 @@ cosmology: !astropy.cosmology.default_cosmology.get
 A_s: 2.1982e-09
 n_s: 0.969453
 redshift: 1
+M_star: !astropy.modeling.models.Linear1D [-0.70798041, -20.37196157]
 growth_function: !skypy.power_spectrum.growth_function
   redshift: $redshift
   cosmology: $cosmology
@@ -23,12 +24,14 @@ tables:
       growth_function: $growth_function
       cosmology: $cosmology
   galaxies:
-    luminosity: !skypy.galaxy.luminosity.herbel_luminosities
-      alpha: -0.5
-      a_m: -0.610
-      b_m: -20.416
-      size: 100
+    magnitude: !skypy.galaxy.luminosity.schechter_lf_magnitude
       redshift: $redshift
+      M_star: $M_star
+      alpha: -0.5
+      m_lim: 35
+      size: 100
+    luminosity: !skypy.utils.astronomy.luminosity_from_absolute_magnitude
+      absolute_magnitude: $galaxies.magnitude
   abundance_matching:
     .init: !skypy.halo.abundance_matching.vale_ostriker
       .depends: [halos.complete, galaxies.complete]
--- a/skypy/galaxy/redshift.py
+++ b/skypy/galaxy/redshift.py
@@ -5,16 +5,21 @@ models.
 """
 
 import numpy as np
-import scipy.special as sc
 import scipy.integrate
+import scipy.special
 
-import skypy.utils.special as special
+from ..utils import uses_default_cosmology, broadcast_arguments, dependent_argument
 
 
 __all__ = [
-    'herbel_redshift',
+    'redshifts_from_comoving_density',
+    'schechter_lf_redshift',
     'smail',
 ]
+
+
+# largest number x such that exp(x) is a float
+_LOGMAX = np.log(np.finfo(0.).max)
 
 
 def smail(z_median, alpha, beta, size=None):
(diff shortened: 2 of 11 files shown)


skypy/galaxy/tests/test_redshift.py
```diff
@@ -1,12 +1,82 @@
...
+    cosmo = FlatLambdaCDM(H0=70, Om0=0.3)
+
+    m_lim = 25.
+    z = np.asarray([1., 2.])
+    M_star = -20
+    phi_star = 1e-3
+    fsky = 1/41253
+
+    # redshifts for blue and red galaxies
+    z_blu = schechter_lf_redshift(z, M_star, phi_star, -1.3, m_lim, fsky, cosmo, noise=False)
```
philipp128, line 21: Typo: "blu' -> "blue"

skypy/galaxy/redshift.py
```diff
@@ -5,243 +5,216 @@
...
-    cdf = cdf / cdf[-1]
-    u = np.random.uniform(size=size)
-    redshift_sample = np.interp(u, cdf, redshift)
+    # integrate density to get expected number of galaxies
+    N = np.trapz(dN_dz, redshift)
 
-    return redshift_sample
+    # Poisson sample galaxy number if requested
+    if noise:
+        N = np.random.poisson(N)
```
philipp128, line 159: When do we want a Poisson sample?

skypy/galaxy/redshift.py
```diff
@@ -5,243 +5,216 @@
...
 
-    Here the luminosity is defined as
+    # compute lower truncation of scaled Schechter random variable
+    lnxmin = m_lim - cosmology.distmod(np.clip(redshift, 1e-10, None)).value
+    lnxmin -= M_star
+    lnxmin *= -0.92103403719761827361
 
-    .. math::
+    # gamma function integrand
+    def f(lnx, a):
```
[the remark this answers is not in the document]
ntessore: The lower limit is given by an apparent magnitude cut now, the numerical integration is only because there is no good general way to compute the upper incomplete gamma function otherwise (that I know of).

skypy/galaxy/redshift.py
```diff
@@ -5,243 +5,216 @@
...
+        return np.exp((a + 1)*lnx - np.exp(lnx)) if lnx < _LOGMAX else 0.
```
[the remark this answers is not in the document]
ntessore: That's just implementation details for computing the upper incomplete gamma function in log space.

skypy/galaxy/redshift.py
```diff
@@ -5,243 +5,216 @@
...
-        \Phi(L, z) = \frac{\Phi_\star(z)}{L_\star(z)}
-            \left(\frac{L}{L_\star(z)}\right)^\alpha
-            \exp\left(-\frac{L}{L_\star(z)}\right) \;.
+    '''
 
-    Here the luminosity is defined as
+    # compute lower truncation of scaled Schechter random variable
+    lnxmin = m_lim - cosmology.distmod(np.clip(redshift, 1e-10, None)).value
+    lnxmin -= M_star
+    lnxmin *= -0.92103403719761827361
```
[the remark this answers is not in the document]
ntessore: It's just the factor to convert magnitudes to `log(x)`?

skypy/galaxy/redshift.py
```diff
@@ -5,243 +5,216 @@
...
-    return redshift_sample
+    # Poisson sample galaxy number if requested
+    if noise:
+        N = np.random.poisson(N)
+    else:
+        N = int(N)
 
+    # cumulative trapezoidal rule to get redshift CDF
+    cdf = dN_dz  # reuse memory
+    np.cumsum((dN_dz[1:]+dN_dz[:-1])/2*np.diff(redshift), out=cdf[1:])
```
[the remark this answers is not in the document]
ntessore: There is no numpy cumulative trapezoidal rule, and I found it over the top to import scipy.integrate for using the same one-liner. I agree that it would be better to have an automatic integration but it's more tricky than one might think, so I would need to think a little about that

skypy/galaxy/tests/test_redshift.py
ntessore: not a typo

skypy/galaxy/redshift.py
ntessore: when we want to simulate shot noise

skypy/galaxy/redshift.py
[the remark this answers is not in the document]
philipp128: Yes. But before we calculated it for only one limit and got the redshift dependence by multiplying with `L_*` and now we also get z dependent limits, right?

skypy/galaxy/redshift.py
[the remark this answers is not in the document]
ntessore: Actually no, in this function you only have the density given at a fixed resolution, so there's not much else you can do.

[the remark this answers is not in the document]
philipp128: And using scipy is not a good alternative?

skypy/galaxy/redshift.py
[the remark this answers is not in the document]
philipp128: Ok got it.

examples/abundance_matching.yml
```diff
@@ -23,12 +24,13 @@ tables:
       growth_function: $growth_function
       cosmology: $cosmology
   galaxies:
-    luminosity: !skypy.galaxy.luminosity.herbel_luminosities
+    magnitude: !skypy.galaxy.luminosity.schechter_lf_magnitude
+      redshift: !numpy.repeat [$redshift, 100]
```
rrjbca, line 28:
```suggestion
      redshift: $redshift
      size: 100
```
I think `schechter_lf_magnitude ` should have an argument `size=None` to handle a case like this where you want to sample multiple luminosities at a fixed redshift. `skypy.utils.random.schechter` already accepts such a size parameter in the correct manner so should be an easy fix.

skypy/galaxy/redshift.py
```diff
@@ -5,243 +5,216 @@
...
-        than high. The default value is 2.0
-    size : int or tuple of ints, optional
-        The number of redshifts to sample. If None one value is sampled.
-    absolute_magnitude_max : float or scalar, optional
-        Upper limit of the considered absolute magnitudes of the galaxies to
-        wanna sample.
-    resolution : int, optional
-        Characterises the resolution of the sampling. Default is 100.
+    redshift : array_like
+        Input redshift range at which the Schechter function parameters are
```
rrjbca, line 40:
```suggestion
        Input redshift grid on which the Schechter function parameters are
```
This confused me for ages as I thought it was expecting a literal range i.e. `redshift=[z_min, z_max]`. Hopefully this makes it clearer

skypy/galaxy/redshift.py
```diff
@@ -5,243 +5,216 @@
...
 
-     References
+    Parameters
     ----------
-    .. [1] Herbel J., Kacprzak T., Amara A. et al., 2017, Journal of Cosmology
-           and Astroparticle Physics, Issue 08, article id. 035 (2017)
+    redshift : array_like
+        Redshifts at which comoving number densities are provided.
+    density : array_like
+        Comoving galaxy number number density at each redshift in Mpc-3.
```
rrjbca, line 123:
```suggestion
        Comoving galaxy number density at each redshift in Mpc-3.
```
Typo

skypy/galaxy/tests/test_luminosity.py
```diff
@@ -1,69 +1,51 @@
...
+    import pytest
 
-def test_calculate_luminosity_star():
-    luminosity_star = lum._calculate_luminosity_star(1, 2.5, 5)
-    assert luminosity_star == 0.001
+    # random cosmology
+    H0 = np.random.uniform(50, 100)
+    Om = np.random.uniform(0.1, 0.5)
+    Ode = np.random.uniform(0.1, 0.9)
+    cosmo = LambdaCDM(H0=H0, Om0=Om, Ode0=Ode)
```
rrjbca, line 14: Any reason to use a random cosmology? Perhaps just use `Planck15`

skypy/galaxy/tests/test_redshift.py
```diff
@@ -1,12 +1,82 @@
...
+
+    # sample with noise this time
+    z_blu = schechter_lf_redshift(z, M_star, phi_star, -1.3, m_lim, fsky, cosmo, noise=True)
+    z_red = schechter_lf_redshift(z, M_star, phi_star, -0.5, m_lim, fsky, cosmo, noise=True)
+
+    # since only two points are given, check uniform distribution between z[0] and z[1]
+    D, p = kstest(z_blu, 'uniform', args=(z[0], z[1]-z[0]))
+    assert p > 0.01, 'D = {}, p = {}'.format(D, p)
+    D, p = kstest(z_red, 'uniform', args=(z[0], z[1]-z[0]))
+    assert p > 0.01, 'D = {}, p = {}'.format(D, p)
```
rrjbca, line 45: Not sure if I like these tests. They only pass because our *implementation* gives a uniform approximation to the true pdf when `len(z)==2`. They don't test against any actual properties of the Schechter function itself e.g. alpha, M_star. And they don't test the implementation if len(z) > 2. But I'm struggling to come up with something better, any thoughts?

rrjbca (changes requested):
Looks great. A few small changes. Also could you move `smail` back to being the first function in `redshift.py` as currently the diff makes it look like this PR is modifying it in a really messy way which will leave a confusing history.

skypy/galaxy/tests/test_redshift.py
ntessore: They are failing now as well, so I am curious about any suggestions, too.

rrjbca (approved):

PR #245, opened 2020-08-05
simplify config file syntax

ntessore:
## Description

Replace the `module` and `function` syntax in the driver with fully qualified `module.function` syntax. All arguments can now be given in the `args` list or dict. Literal strings must be escaped by an initial `~` (tilde).

## Checklist
- [x] Write unit tests


skypy/pipeline/driver.py
```diff
@@ -102,20 +110,23 @@ def isfunction(f):
                 dag.add_node(job)
 
         # Add edges for all requirements and dependencies
+        def deps(settings):
+            deps = settings.get('depends', [])
+            args = settings.get('args', [])
+            for k, v in _items(args):
+                if isinstance(v, str) and v[0] != '~':
+                    deps.append(v)
+            return deps
```
rrjbca, line 119:
```suggestion
        def deps(settings, dag):
            args = settings.get('args', [])
            deps = settings.get('depends', [])
            deps += [v for _, v in _items(args) if v in dag]
            return deps
```
What do you think about this alternative? Instead of requiring an escape character for string literals, just assume that if a string is in the graph of jobs then it is a dependency, otherwise assume it is a string argument. The config files would probably be cleaner but it limits the choice of variable names so that they don't clash with any string arguments. Not sure which has preferable behaviour if the user typos a variable name etc.

skypy/pipeline/driver.py
```diff
@@ -138,27 +149,38 @@ def isfunction(f):
...
+        function = getattr(module, function_path[-1])
 
         # Parse arguments
         args = config.get('args', [])
-        kwargs = config.get('kwargs', {})
-        req = config.get('requires', {})
-        req = {k: self.__getitem__(v) for k, v in req.items()}
+        for k, v in _items(args):
+            if isinstance(v, str):
+                args[k] = self[v]
```
rrjbca, line 169:
```suggestion
        for k, v in _items(args):
            if isinstance(v, str):
                if v[0] in '"\'' and v[0] == v[-1]:
                    # v is a literal string inside quotes; remove the quotes
                    args[k] = v[1:-1]
                else:
                    # v is the name of a variable
                    args[k] = self[v]
```

skypy/pipeline/driver.py
```diff
@@ -138,27 +149,38 @@ def isfunction(f):
...
+        else:
+            result = function(*args)
+
+        return result
 
     def __getitem__(self, label):
+        # do not parse literal strings
+        if label[0] in '"\'' and label[0] == label[-1]:
+            return label[1:-1]
+
```
rrjbca, line 183:
```suggestion
```

skypy/pipeline/driver.py
```diff
@@ -138,25 +159,36 @@ def isfunction(f):
...
         args = config.get('args', [])
-        kwargs = config.get('kwargs', {})
-        req = config.get('requires', {})
-        req = {k: self.__getitem__(v) for k, v in req.items()}
+        for k, v in _items(args):
+            if isinstance(v, str):
+                if _is_string_argument(v):
+                    args[k] = _parse_string_argument(v)
+                else:
+                    # get column
```
rrjbca, line 182:
```suggestion
                    # get variable or table column
```

rrjbca (approved):

PR #249, opened 2020-08-19
Class

### LINKED ISSUE #21: Integration with CLASS
## Description
This model computes the linear power spectrum as a function of wavenumbers, k, and redshift. The model uses CLASS. 
"The purpose of CLASS is to simulate the evolution of linear perturbations in the universe and to compute CMB and large scale structure observables. Its name also comes from the fact that it is written in object-oriented style mimicking the notion of class. Classes are a wonderfull programming feature available e.g. in C++ and python, but these languages are known to be less vectorizable/parallelizable than plain C (or Fortran), and hence potentially slower. For CLASS we choose to use plain C for high performances, while organizing the code in a few modules that reproduce the architecture and philosophy of C++ classes, for optimal readability and modularity."
 
## Inputs
 - Background cosmology

## Outputs
 - The linear power spectrum
 - Wavenumber
 - Redshift

## References
- https://lesgourg.github.io/class_public/class.html

## Additional
Need to consider the choice of distance units; do we use Mpc or Mpc/h?

itrharrison:
## Description

Allows computation of the the linear matter power spectrum with CLASS instead of CAMB.

Merging will close #21 

## Checklist
- [X] Write unit tests


ntessore (changes requested):
If there is a pip-installable package for CLASS, could you add it to our [setup.cfg](https://github.com/skypyproject/skypy/blob/f069858a6c9882a02c09131967070ec3e864e8ae/setup.cfg#L33-L34) so that the CLASS test can run and report coverage?

rrjbca:
Perhaps try [`classylss`](https://pypi.org/project/classylss/)?

itrharrison:
@rrjbca okay, I changed this so it now works with the `classylss` pip package. I think it is wrong in some of its documentation about little h units. I will put together a notebook hopefully showing why.

ntessore:
I am not sure how good of a choice it is to base our integration with CLASS on the unofficial `classylss` package if CLASS itself comes with official python bindings. There's the convenience of having `classylss` pip-installable vs the fact that it could break at any time.

skypy/power_spectrum/_classy.py
```diff
@@ -0,0 +1,95 @@
...
+    >>> redshift = np.array([0, 1])
+    >>> wavenumber = np.array([1.e-2, 1.e-1, 1e0])
+    >>> A_s = 2.e-9
+    >>> n_s = 0.965
+    >>> z_reio = 10.
+    >>> classy(wavenumber, redshift, cosmology, A_s, n_s, z_reio)  # doctest: +SKIP
+    array([[2.34758952e+04, 8.70837957e+03],
+           [3.03660813e+03, 1.12836115e+03],
+           [2.53124880e+01, 9.40802814e+00]])
+    References
```
ntessore, line 54:
```suggestion

    References
```

skypy/power_spectrum/_classy.py
```diff
@@ -0,0 +1,95 @@
...
+    doi : 10.1088/1475-7516/2011/07/034
+    arXiv: 1104.2932, 1104.2933
+
+    """
+    try:
+        from classy import Class
+    except ImportError:
+        raise Exception("classy is required to use skypy.linear.classy")
+
+    redshift = np.atleast_1d(redshift)
```
ntessore, line 65: Here you are saying at least 1d, but below the array is set over the first dimension only. I think there's a better approach, see below.

skypy/power_spectrum/_classy.py
```diff
@@ -0,0 +1,95 @@
...
+
+    k = wavenumber * (1. / u.Mpc)
+
```
ntessore, line 68: I think the units are never actually used?
```suggestion
```

skypy/power_spectrum/_classy.py
```diff
@@ -0,0 +1,95 @@
...
+    h2 = cosmology.h * cosmology.h
+
+    params = {
+        'output': 'mPk',
+        'P_k_max_1/Mpc':  k.max().value,
```
ntessore, line 73:
```suggestion
        'P_k_max_1/Mpc':  np.max(wavenumber),
```

skypy/power_spectrum/_classy.py
```diff
@@ -0,0 +1,95 @@
...
+    classy_obj.set(params)
+    classy_obj.compute()
+
+    pzk = np.zeros([redshift.shape[0], wavenumber.shape[0]])
+
+    for ik, k in enumerate(k.value):
+        for iz, z in enumerate(redshift):
+            pzk[iz, ik] = classy_obj.pk_lin(k, z)
+
+    return pzk.T
```
ntessore, line 95: I think it makes more sense to accept any shape and just make it so that the first dimensions of the output correspond to redshift, and the last dimensions correspond to wavenumber:
```suggestion
    z = np.expand_dims(redshift, (-1,)*np.ndim(wavenumber))
    k = np.expand_dims(wavenumber, (0,)*np.ndim(redshift))
    z, k = np.broadcast_arrays(z, k)
    pzk = np.empty(z.shape)

    for i in np.ndindex(*pzk.shape):
        pzk[i] = classy_obj.pk_lin(k[i], z[i])

    if pzk.ndim == 0:
        pzk = pzk.item()

    return pzk
```

tox.ini
```diff
@@ -77,6 +77,8 @@ deps =
     oldest: numpy==1.16.*
     oldest: scipy==1.2.*
 
+    all: git+https://github.com/ntessore/classy.git@v2.8.1#egg=classy
```
ntessore, line 80: This has now been moved:
```suggestion
    all: git+https://github.com/skypyproject/classy.git@v2.8.1#egg=classy
```

ntessore: The better way to specify the dependency is the following change to `setup.cfg`:
```diff
 [options.extras_require]
 test =
     pytest-astropy
 all =
     camb
+    classy @ git+https://github.com/skypyproject/classy.git@v2.8.1#egg=classy
 docs =
     sphinx-astropy
     matplotlib
```

ntessore (dismissed):

skypy/power_spectrum/_classy.py
```diff
@@ -0,0 +1,96 @@
...
+        'z_pk': ', '.join(str(z) for z in redshift),
+        'A_s':       A_s,
+        'n_s':       n_s,
+        'H0':        cosmology.H0.value,
+        'omega_b':   cosmology.Ob0 * h2,
+        'omega_cdm': cosmology.Odm0 * h2,
+        'T_cmb':     cosmology.Tcmb0.value,
+        'N_eff':     cosmology.Neff,
+        'z_reio':    z_reio,
+    }
```
rrjbca, line 79:
```suggestion
    params = {
        'P_k_max_1/Mpc': np.max(wavenumber),
        'z_pk': ', '.join(str(z) for z in redshift),
        'H0': cosmology.H0.value,
        'omega_b': cosmology.Ob0 * h2,
        'omega_cdm': cosmology.Odm0 * h2,
        'T_cmb': cosmology.Tcmb0.value,
        'N_eff': cosmology.Neff,
    }
    
    params.update(kwargs)
```
- It makes sense to only explicitly configure quantities depending on `wavenumber`, `redshift` and `cosmology` here due to their different interface; take everything else from `kwargs`.
- Are there any other parameters from `cosmology` that we should add here?
- Do we need to restrict ourselves to only the matter power spectrum? We could create a wrapper function `class_matter_power_spectrum` that sets `output = 'mPk'` if that's convenient? 
- Whitespace

rrjbca (changes requested):

itrharrison:
@rrjbca I did not implement your suggestion of removing 'mPk' as a parameter to classy. The modules within this section are all explicitly matter power spectrum modules, so e.g. CMB TT spectra should be computed else where. It might be worth thinking about whether we would want to refactor things though, as it would also seem incorrect to split up the calculations of mPk and ClTT when they are both done internally to wrapped camb/class etc.

skypy/power_spectrum/_classy.py
```diff
@@ -0,0 +1,94 @@
...
+        spectrum.
+    cosmology : astropy.cosmology.Cosmology
+        Cosmology object providing omega_matter, omega_baryon, Hubble
+        parameter and CMB temperature in the present day
+    A_s : float
+        Cosmology parameter, amplitude normalisation of curvature perturbation
+        power spectrum
+    n_s : float
+        Cosmology parameter, spectral index of scalar perturbation power
+        spectrum
```
rrjbca, line 28:
```suggestion
```

skypy/power_spectrum/_classy.py
```diff
@@ -0,0 +1,94 @@
+import numpy as np
+
+__all__ = [
+    'classy',
+]
+
+
+def classy(wavenumber, redshift, cosmology, **kwargs):
+    """ Return the CLASS computation of the linear matter power spectrum, on a
+    two dimensional grid of wavenumber and redshift
```
rrjbca, line 10:
```suggestion
    """ Return the CLASS computation of the linear matter power spectrum, on a
    two dimensional grid of wavenumber and redshift.
    
    Additional CLASS parameters can be passed via keyword arguments. 
```

skypy/power_spectrum/_classy.py
```diff
@@ -0,0 +1,94 @@
...
+    z = np.expand_dims(redshift, (-1,)*np.ndim(wavenumber))
+    k = np.expand_dims(wavenumber, (0,)*np.ndim(redshift))
+    z, k = np.broadcast_arrays(z, k)
+    pzk = np.empty(z.shape)
+
+    for i in np.ndindex(*pzk.shape):
+        pzk[i] = classy_obj.pk_lin(k[i], z[i])
+
+    if pzk.ndim == 0:
+        pzk = pzk.item()  # pragma: no cover
```
rrjbca, line 92:
```suggestion
        pzk = pzk.item()
```
Don't skip coverage checks here. Instead write a quick test where z and k are both scalars and check that the return value is a scalar.

skypy/power_spectrum/_classy.py
```diff
@@ -0,0 +1,94 @@
...
+        from classy import Class
+    except ImportError:
+        raise Exception("classy is required to use skypy.linear.classy")
+
+    h2 = cosmology.h * cosmology.h
+
+    params = {
+        'output': 'mPk',
+        'P_k_max_1/Mpc':  np.max(wavenumber),
+        'z_pk': ', '.join(str(z) for z in redshift),
```
rrjbca, line 69:
```suggestion
        'z_pk': ', '.join(str(z) for z in np.atleast_1d(redshift)),
```
Something like this is probably needed to support scalar redshift

rrjbca (changes requested):
Looks good just a few final changes...

rrjbca (approved):

PR #252, opened 2020-08-24
Remove little h units from CAMB

### LINKED ISSUE #250: Wrong units for CAMB P(k)
**Describe the bug**
We currently return the output from the CAMB `get_matter_power_spectrum` function and assume these do not come with $h$ units. However, the output from this function does include those units.

We should instead use `get_matter_power_interpolator` which creates a spline interpolation for the matter power spectrum, and has a option to remove $h$ units. See below for this, and a comparison with the output from CLASS.

camb_class_comparison

itrharrison:
## Description

Switch to the CAMB matter power interpolator and remove the return units of $h^3$.

Note that I have also added an argument for reionisation redshift and forced the neutrino mass to zero. Both of these are in order that the CAMB output can be made to agree with CLASS (#249) (it should be possible to make them agree with varying these two parameters, but I haven't worked out how yet).

Merging will close #250 .

## Checklist
- [X] Write unit tests


itrharrison:
The agreement between CAMB and CLASS with neutrino mass fixed to zero is shown here:
camb_class_comparison

ntessore:
Would CAMB and CLASS agree if a cosmology with zero neutrino mass is passed?

skypy/power_spectrum/_camb.py
```diff
@@ -72,29 +77,31 @@ def camb(wavenumber, redshift, cosmology, A_s, n_s):
                        omch2=cosmology.Odm0 * h2,
                        omk=cosmology.Ok0,
                        TCMB=cosmology.Tcmb0.value,
-                       mnu=np.sum(cosmology.m_nu.value),
+                       mnu=0.,  # for now, need to solve agreement with CLASS
```
ntessore, line 80: Keep the neutrino mass here (but use explicit neutrino mass units). Agreement with CLASS is for cosmologies with massless neutrinos.

```suggestion
                       mnu=np.sum(cosmology.m_nu.to_value(u.eV)),
```

skypy/power_spectrum/_camb.py
```diff
@@ -72,29 +77,31 @@ def camb(wavenumber, redshift, cosmology, A_s, n_s):
...
                        standard_neutrino_neff=cosmology.Neff
                        )
 
-    # camb requires redshifts to be in decreasing order
-    redshift_order = np.argsort(redshift)[::-1]
+    # camb interpolator requires redshifts to be in increasing order
+    redshift_order = np.argsort(redshift)
```
ntessore, line 86: CAMB interpolator makes no assumptions about redshift order as far as I can tell.
```suggestion
```

skypy/power_spectrum/_camb.py
```diff
@@ -72,29 +77,31 @@ def camb(wavenumber, redshift, cosmology, A_s, n_s):
...
+    pars.Reion.redshift = z_reio
+
     pars.set_matter_power(redshifts=list(redshift[redshift_order]),
                           kmax=np.max(wavenumber))
 
     pars.NonLinear = model.NonLinear_none
 
-    results = get_results(pars)
-
     k = wavenumber * (1. / u.Mpc)
```
ntessore, line 98: This is no longer necessary, `wavenumber` is already the right array.
```suggestion
```

skypy/power_spectrum/_camb.py
```diff
@@ -72,29 +77,31 @@ def camb(wavenumber, redshift, cosmology, A_s, n_s):
...
 
-    k_h = k.to((u.littleh / u.Mpc), u.with_H0(cosmology.H0))
+    pk_interp = get_matter_power_interpolator(pars,
+                                              nonlinear=False,
+                                              hubble_units=False, k_hunit=False,
+                                              kmax=np.max(k.value),
+                                              zmax=redshift.max())
```
ntessore, line 103: The input arrays are ok for this. Use `np.max()` because they might not be numpy arrays.
```suggestion
                                              kmax=np.max(wavenumber),
                                              zmax=np.max(redshift))
```

skypy/power_spectrum/_camb.py
```diff
@@ -72,29 +77,31 @@ def camb(wavenumber, redshift, cosmology, A_s, n_s):
...
 
-    kh, z, pzk = results.get_matter_power_spectrum(minkh=np.min(k_h.value),
-                                                   maxkh=np.max(k_h.value),
-                                                   npoints=len(k_h.value))
+    pzk = pk_interp.P(redshift[redshift_order], k.value)
```
ntessore, line 105: Use arrays as provided.
```suggestion
    pzk = pk_interp.P(redshift, wavenumber)
```

skypy/power_spectrum/_camb.py
```diff
@@ -72,29 +77,31 @@ def camb(wavenumber, redshift, cosmology, A_s, n_s):
...
 
-    return pzk[redshift_order[::-1]].reshape(return_shape)
+    return pzk[redshift_order].reshape(return_shape)
```
ntessore, line 107:
```suggestion
    return pzk.reshape(return_shape)
```

skypy/power_spectrum/_camb.py
```diff
@@ -7,10 +7,12 @@
 ]
 
 
-def camb(wavenumber, redshift, cosmology, A_s, n_s):
+def camb(wavenumber, redshift, cosmology,
+         A_s, n_s, z_reio):
     r'''CAMB linear matter power spectrum.
     Compute the linear matter power spectrum on a two dimensional grid of
     redshift and wavenumber using CAMB [1]_.
+    Currently assumes massless neutrinos.
```
ntessore, line 15: We shouldn't assume this.
```suggestion
```

skypy/power_spectrum/_camb.py
```diff
@@ -72,29 +77,31 @@ def camb(wavenumber, redshift, cosmology, A_s, n_s):
...
+    # camb interpolator requires redshifts to be in increasing order
+    redshift_order = np.argsort(redshift)
 
     pars.InitPower.ns = n_s
     pars.InitPower.As = A_s
 
+    pars.Reion.redshift = z_reio
+
     pars.set_matter_power(redshifts=list(redshift[redshift_order]),
                           kmax=np.max(wavenumber))
```
ntessore, line 94: No need to set the parameters outside of the interpolator call.
```suggestion
```

ntessore (changes requested):
I think with the use of the interpolator, the code can be made a lot more compact.

skypy/power_spectrum/_camb.py
```diff
@@ -72,29 +76,26 @@ def camb(wavenumber, redshift, cosmology, A_s, n_s):
...
                        TCMB=cosmology.Tcmb0.value,
-                       mnu=np.sum(cosmology.m_nu.value),
+                       mnu=np.sum(cosmology.m_nu.to_value(u.eV)),
                        standard_neutrino_neff=cosmology.Neff
                        )
 
-    # camb requires redshifts to be in decreasing order
-    redshift_order = np.argsort(redshift)[::-1]
+    # camb interpolator requires redshifts to be in increasing order
+    redshift_order = np.argsort(redshift)
```
ntessore, line 84: In complete opposition to the rubbish I was saying before...
```suggestion
    # camb interpolator requires redshifts and wavenumber to be in increasing order
    redshift_order = np.argsort(redshift)
    wavenumber_order = np.argsort(wavenumber)
```
... and then below the requisite change as well.

rrjbca (changes requested):
The re-ionisation redshift feature should be moved to a separate pull request. Since the little h factor is a bug fix, we may want to backport it and would need to be able to do so independent of the API-breaking re-ionisation feature. Combining them will also create a confusing log that implies the re-ionisation redshift was required to fix this bug which it isn't.

Also I dislike modifying the unit test from `Planck15 ` to `Pl15massless `. Since fixing the h factor bug doesn't modify the API it is only be necessary to update the values in `camb_result.txt`.

skypy/power_spectrum/_camb.py
```diff
@@ -72,29 +76,26 @@ def camb(wavenumber, redshift, cosmology, A_s, n_s):
...
 
-    results = get_results(pars)
+    pk_interp = get_matter_power_interpolator(pars,
+                                              nonlinear=False,
+                                              hubble_units=False, k_hunit=False,
+                                              kmax=np.max(wavenumber),
+                                              zmax=np.max(redshift))
 
-    k = wavenumber * (1. / u.Mpc)
+    pzk = pk_interp.P(redshift[redshift_order], wavenumber)
```
ntessore, line 99: This is the matching change:
```suggestion
    pzk = pk_interp.P(redshift[redshift_order], wavenumber[wavenumber_order])
```

skypy/power_spectrum/_camb.py
```diff
@@ -72,29 +76,26 @@ def camb(wavenumber, redshift, cosmology, A_s, n_s):
...
 
-    k_h = k.to((u.littleh / u.Mpc), u.with_H0(cosmology.H0))
-
-    kh, z, pzk = results.get_matter_power_spectrum(minkh=np.min(k_h.value),
-                                                   maxkh=np.max(k_h.value),
-                                                   npoints=len(k_h.value))
-
-    return pzk[redshift_order[::-1]].reshape(return_shape)
+    return pzk[redshift_order].reshape(return_shape)
```
ntessore, line 101: I don't quite understand how the same order can index the return array: if `z = [1.9, 1.5, 1.7]` then its argsort is `a = [1, 2, 0]`. The computed array has values `[p(1.5), p(1.7), p(1.9)]` which must be sorted by `[2, 0, 1]`, i.e. the inverse of the original argsort array: `a[a] = [2, 0, 1]`.

skypy/power_spectrum/_camb.py
```diff
@@ -29,6 +30,8 @@ def camb(wavenumber, redshift, cosmology, A_s, n_s):
     n_s : float
```
Lucia-Fonseca, line 30: Does it make sense to make some changes to the docstring in this PR?

skypy/power_spectrum/_camb.py
```diff
@@ -7,7 +7,8 @@
 ]
```
Lucia-Fonseca, line 8: Let me suggest `from astropy import units`

skypy/power_spectrum/_camb.py
```diff
@@ -72,29 +76,26 @@ def camb(wavenumber, redshift, cosmology, A_s, n_s):
                        omch2=cosmology.Odm0 * h2,
                        omk=cosmology.Ok0,
                        TCMB=cosmology.Tcmb0.value,
-                       mnu=np.sum(cosmology.m_nu.value),
+                       mnu=np.sum(cosmology.m_nu.to_value(u.eV)),
```
Lucia-Fonseca, line 79:
```suggestion
                       mnu=np.sum(cosmology.m_nu.to_value(units.eV)),
```

skypy/power_spectrum/tests/test_camb.py
```diff
@@ -24,15 +25,17 @@ def test_camb():
     '''
     from skypy.power_spectrum import camb
 
+    Pl15massless = Planck15.clone(name='Planck 15 massless neutrino', m_nu=[0., 0., 0.]*u.eV)
```
Lucia-Fonseca, line 28:
```suggestion
    Pl15massless = Planck15.clone(name='Planck 15 massless neutrino', m_nu=[0., 0., 0.]*units.eV)
```

Lucia-Fonseca (changes requested):
I agree, it is confusing to modify the code to add a new feature when the PR is to address a typo.

skypy/power_spectrum/_camb.py
ntessore: I disagree: `from astropy import units as u` is a prevalent idiom that is used everywhere in astropy, because `units.eV` is harder to parse than `u.eV`.

Lucia-Fonseca: I agree. However, I thought we agreed on avoiding this type of abbreviation?

ntessore: I wasn’t part of any discussion on this so I don’t know. I agree to not coming up with unconventional ones like `import subprocess as sbp` but I think the default ones like `np` and `u` are helpful.

rrjbca: I have given reviews before suggesting not to use `import a as b` except for `import numpy as np` as I think that is the only one that can reasonably be considered a "convention" in the community. Astropy may use `import astropy.units as u` internally but that doesn't mean your average astronomer reading python code would know what `u` is without checking the imports. And I disagree I think `units.eV` is ***much*** clearer than `u.eV`. In pretty much all cases I feel it's not worth the loss of clarity to abbreviate names especially now that we allow more than 80 characters per line.

skypy/power_spectrum/_camb.py
itrharrison: Can you say a little bit more what you mean about this?

rrjbca: I don't see any problem other than the units for the wavenumber should be left as `Mpc-1` not changed to `[Mpc^-1]` for consistency with #272

Lucia-Fonseca: Agreed with Richard

skypy/power_spectrum/_camb.py
```diff
@@ -72,29 +76,27 @@ def camb(wavenumber, redshift, cosmology, A_s, n_s):
...
+    pzk = pk_interp.P(redshift[redshift_order], wavenumber[wavenumber_order])
 
-    k_h = k.to((u.littleh / u.Mpc), u.with_H0(cosmology.H0))
-
-    kh, z, pzk = results.get_matter_power_spectrum(minkh=np.min(k_h.value),
-                                                   maxkh=np.max(k_h.value),
-                                                   npoints=len(k_h.value))
-
-    return pzk[redshift_order[::-1]].reshape(return_shape)
+    return pzk[redshift_order].reshape(return_shape)
```
Lucia-Fonseca, line 102: It works well for me

Lucia-Fonseca (changes requested):
Looks good to me.
Just one change in units from `[Mpc^-1]` to `Mpc-1`.

skypy/power_spectrum/_camb.py
```diff
@@ -1,13 +1,14 @@
...
 
 
 __all__ = [
     'camb',
 ]
 
 
-def camb(wavenumber, redshift, cosmology, A_s, n_s):
+def camb(wavenumber, redshift, cosmology,
+         A_s, n_s):
```
rrjbca, line 11:
```suggestion
def camb(wavenumber, redshift, cosmology, A_s, n_s):
```

skypy/power_spectrum/_camb.py
```diff
@@ -47,7 +48,8 @@ def camb(wavenumber, redshift, cosmology, A_s, n_s):
     >>> wavenumber = np.array([1.e-2, 1.e-1, 1e0])
     >>> A_s = 2.e-9
     >>> n_s = 0.965
-    >>> power_spectrum = camb(wavenumber, redshift, cosmology, A_s, n_s) # doctest: +SKIP
+    >>> z_reio = 10.
+    >>> power_spectrum = camb(wavenumber, redshift, cosmology, A_s, n_s, z_reio) # doctest: +SKIP
```
rrjbca, line 52:
```suggestion
    >>> power_spectrum = camb(wavenumber, redshift, cosmology, A_s, n_s) # doctest: +SKIP
```

skypy/power_spectrum/tests/test_camb.py
```diff
@@ -1,12 +1,14 @@
 import numpy as np
 from astropy.cosmology import Planck15
+from astropy import units
 from astropy.units import allclose
 from astropy.utils.data import get_pkg_data_filename
 import pytest
+import pdb
```
rrjbca, line 7:
```suggestion
```

skypy/power_spectrum/tests/test_camb.py
```diff
@@ -24,15 +26,17 @@ def test_camb():
     '''
     from skypy.power_spectrum import camb
 
+    Pl15massless = Planck15.clone(name='Planck 15 massless neutrino', m_nu=[0., 0., 0.]*units.eV)
+
```
rrjbca, line 30:
```suggestion
```

skypy/power_spectrum/tests/test_camb.py
```diff
@@ -24,15 +26,17 @@ def test_camb():
...
     # test shape and compare with the mocked power spectrum
     redshift = [0.0, 1.0]
     wavenumber = np.logspace(-4.0, np.log10(2.0), 200)
-    pkz = camb(wavenumber, redshift, Planck15, 2.e-9, 0.965)
-    assert pkz.shape == (len(redshift), len(wavenumber))
-    assert allclose(pkz, test_pkz, rtol=1.e-4)
+    pzk = camb(wavenumber, redshift, Pl15massless, 2.e-9, 0.965)
+    assert pzk.shape == (len(redshift), len(wavenumber))
+    assert allclose(pzk, test_pzk, rtol=1.e-4)
```
rrjbca, line 36:
```suggestion
    pzk = camb(wavenumber, redshift, Planck15, 2.e-9, 0.965)
    assert pzk.shape == (len(redshift), len(wavenumber))
    assert allclose(pzk, test_pzk, rtol=1.e-4)
```
Revert to using Planck15 in the camb tests. You will still need to update camb_results.txt

skypy/power_spectrum/tests/test_class.py
```diff
@@ -25,22 +25,20 @@ def test_classy():
     '''
     from skypy.power_spectrum import classy
 
-    Pl15massless = Planck15.clone(name='Planck 15 massless neutrino', m_nu=[0., 0., 0.]*u.eV)
-
     redshift = [0.0, 1.0]
     wavenumber = np.logspace(-4.0, np.log10(2.0), 200)
-    pzk = classy(wavenumber, redshift, Pl15massless)
+    pzk = classy(wavenumber, redshift, Planck15)
```
rrjbca, line 30:
```suggestion
    pzk = classy(wavenumber, redshift, Pl15massless)
```
Don't make any changes to the class tests in this PR. Revert to using Pl15massless cosmology. It doesn't matter that this is not the same as what we test in camb

rrjbca (changes requested):

rrjbca (approved):

ntessore (approved):

Lucia-Fonseca (approved):
Looks good!

PR #255, opened 2020-08-27
simplify config file syntax

ntessore:
## Description

This PR makes use of YAML tags (`!tag`) to define functions and references in config files in a non-hackish way.

```yaml
cosmology: !astropy.cosmology.default_cosmology.get
tables:
  merger_rates:
    redshift: !skypy.galaxy.redshift.herbel_redshift
      alpha: -0.5
      a_phi: -2.232
      b_phi: 0.0141
      a_m: -0.610
      b_m: -20.416
      size: 1000
      cosmology: !cosmology
    luminosity: !skypy.galaxy.luminosity.herbel_luminosities
      alpha: -0.5
      a_m: -0.610
      b_m: -20.416
      redshift: !merger_rates.redshift
    ns_ns_rate: !skypy.gravitational_wave.b_band_merger_rate
      population: NS-NS
      optimism: low
      luminosity: !merger_rates.luminosity
```

## Checklist
- [x] Write unit tests


skypy/pipeline/driver.py
```diff
@@ -103,7 +103,7 @@ def execute(self, configuration, file_format=None, overwrite=False):
 
         # Variables initialised by value don't require function evaluations
         def isfunction(f):
-            return isinstance(f, Mapping) and 'function' in f
+            return isinstance(f, Mapping)
         variables = {k: v for k, v in config.items() if not isfunction(v)}
```
rrjbca, line 107:
```suggestion
        variables = {k: v for k, v in config.items() if not isinstance(v, Mapping)}
```
This is effectively what is happening now. Does this mean that we can't define dictionaries as variables? Is this an issue?

rrjbca:
Could you also update the docstrings for SkyPyDriver.execute and skypy/pipeline/scripts/skypy.py to reflect the new sytax

skypy/pipeline/driver.py
ntessore: This is an issue, which is essentially the question where variables or functions can appear. I think the logical next step from this change would be to have columns that support multiple function calls,
```yaml
my_column:
- my_module.my_sequential_function_1:
    arg1: ...
    arg2: ...
- my_module.my_sequential_function_2:
    arg3: ...
    arg4: ...
```
which would bring the same issue with lists and variables as lists (more common than dicts perhaps). So if there can be better specified rules for where variables can appear (can columns be variables?) that would resolve both situations.

ntessore: So this is strings vs. variable names all over again.

rrjbca: We could have a config section `variables` where all entires are assigned their values as parsed by `pyyaml`. That would also conveniently get around the need for escaping literal strings as you could just define them all as variables and then assume any string argument for a function is actually a variable name. Then perhaps have another section for "variables" but that are calculated from functions:
```
cosmology:
    astropy.cosmology.default_cosmology.get: []

variables:
    int_variable: 1
    float_variable: 2.3
    string_variable: A literal string
    list_variable:
    - element1
    - element2
    dict_variable:
        key1: value1
        key2: value2

variables_from_functions: # Obviously needs a better name
    two_squared:
        np.square: [2]
    gamma_one:
        scipy.special.gamma: [int_variable]

tables:
    ....
```

ntessore: I think I found a rather neat solution. Literal strings are tagged with `!str` and literal dicts are tagged with `!dict`. That can be expanded to any other type we want to support in the future.

skypy/pipeline/tests/test_driver.py
```diff
@@ -90,29 +82,40 @@ def test_driver():
...
-    config = {'test_func': {
-                'function': 'list',
-                'args': [
-                    '"hello world"']},
-              'test_func2': {
-                'function': 'len',
-                'args': [
-                    'test_func']}}
+    config = {'test_func': ('list', 'hello world'),
+              'test_func2': ('len', 'test_func')}
```
rrjbca, line 103:
```suggestion
              'test_func2': ('len', ('test_func',))}
```
This is supposed to be a reference to the variable `test_func` not the literal string `"test_func"`. I think this is the correct new syntax for that. Hence the test below should remains as `assert driver.test_func2 == 11`

ntessore: Hah I couldn't figure out why it was 11 🤡

skypy/pipeline/driver.py
```diff
@@ -173,24 +164,62 @@ def _call_from_config(self, config):
...
+            # get explicit dependencies
+            deps = args.pop('.depends', [])
+            # turn a single value into a list
+            if isinstance(deps, str) or not isinstance(deps, list):
+                deps = [deps]
+            # recurse remaining kwargs
+            return deps + sum([self.get_deps(a) for a in args.values()], [])
+        elif isinstance(args, list):
+            # recurse args
+            return sum([self.get_deps(a) for a in args], [])
```
rrjbca, line 217:
```suggestion
            return deps + [self.get_deps(a) for a in args.values()]
        elif isinstance(args, list):
            # recurse args
            return [self.get_deps(a) for a in args]
```
Not sure if/why sum is needed here? Otherwise I think this is ready to merge

ntessore: We want the flat union of the lists, not nested lists.

rrjbca: Ah right because `get_deps(a)` returns a list. Got it. I'll approve and merge now

rrjbca (approved):

PR #261, opened 2020-09-03
Vale & Ostriker 2004 halo-galaxy model

rrjbca:
## Description
Replace the existing `skypy.halo.abundance_matching.vale_ostriker` function with the complete model from Vale & Ostriker 2004:
- sample dark matter halos from a Press-Schechter mass function
- sample subhalo masses from a Press-Schechter mass function with a fixed mass fraction in subhalos,  fixed stripping factor and maximum mass given by half the parent mass
- sample galaxy magnitudes from a Schechter luminosity function
- match galaxies with (sub)halos via a monotonic mass-luminosity relation

## Checklist
- [x] Write unit tests


skypy/halo/abundance_matching.py
```diff
@@ -44,6 +46,13 @@ def vale_ostriker(halos, galaxies, mass='mass', luminosity='luminosity',
     .. [1] Vale A., Ostriker J. P., 2004, MNRAS, 353, 189
     """
 
+    if subhalos:
+        if mass not in halos.columns:
+            raise ValueError("{} is not a column in halos".format(mass))
+        if mass not in subhalos.columns:
+            raise ValueError("{} is not a column in subhalos".format(mass))
+        halos = vstack((halos, subhalos))
+
```
ntessore, line 55: What if my halo mass column is `halo_mass` and my subhalo mass column is `subhalo_mass`? Wouldn't it be much easier to pass columns/arrays instead of tables?

rrjbca: Yes perhaps. The restriction then would be that you couldn't trivially "carry over" any other existing columns in the (sub)halo / galaxy tables into the new matched table.

ntessore: In that case, wouldn't it make more sense to have the user call `astropy.table.vstack` explicitly on the halos and subhalos tables, since they need to have matching definitions for the stacking to work?

rrjbca: For the user to have to call vstack as part of their pipeline is perhaps a bit unintuitive? I like the idea of wrapping it in an abundance matching function in skypy, although perhaps this PR isn't the best design. I think you mentioned before the idea of incorporating the entire Vale & Ostriker 2004 model into a single function, which would be the other extreme.

skypy/halo/tests/test_abundance_matching.py
```diff
@@ -39,3 +39,19 @@ def test_vale_ostriker():
...
+        vale_ostriker(s, g, subhalos=h)
+
+    # Test joining halos and subhalos.
+    nh, ng, ns = 10, 15, 20
+    h = Table(data=np.random.uniform(1, 2, size=(nh, 1)), names=['mass'])
+    g = Table(data=np.random.uniform(size=(ng, 1)), names=['luminosity'])
+    s = Table(data=np.random.uniform(size=(ns, 1)), names=['mass'])
+    matched_subhalos = vale_ostriker(h, g, subhalos=s)
+    assert sorted(matched_subhalos['mass'])[:-nh] == sorted(s['mass'])[nh-ng:]
+    assert sorted(matched_subhalos['mass'])[-nh:] == sorted(h['mass'])
```
Lucia-Fonseca, line 57: This works for me! Just a question:
I would expect `ng` and `ns` to be equal (every halo and subhalo hosting a galaxy).  Have you chosen these numbers for any particular reason?

skypy/halo/abundance_matching.py
Lucia-Fonseca: Do we need to keep track of which subhalos correspond to a parent halo here?

Lucia-Fonseca (changes requested):
Works well for me!
Perhaps an example section is missing.

skypy/halo/abundance_matching.py
```diff
@@ -14,36 +14,75 @@
...
+    # Sample halo and subhalo masses
+    halo_mass = press_schechter(**halo_kwargs)
+    halo_mass[::-1].sort()
+    n_subhalos = number_subhalos(halo_mass, **subhalo_kwargs)
+    del(subhalo_kwargs['gamma_M'])
+    del(subhalo_kwargs['noise'])
+    subhalo_mass = subhalo_mass_sampler(halo_mass, n_subhalos, **subhalo_kwargs)
+
+    # Assign subhalos to groups with their parent halos
+    n_halos = halo_kwargs.get('size')
```
ntessore, line 65:
```suggestion
    n_halos = len(halos)
```

skypy/halo/abundance_matching.py
```diff
@@ -14,36 +14,75 @@
...
+        subhalo_group[first:last] = id
+
+    # Concatenate halos and subhalos
+    mass = np.concatenate([halo_mass, subhalo_mass])
+    group = np.concatenate([halo_group, subhalo_group])
+    parent = np.array((True,) * n_halos + (False,) * total_subhalos)
+
+    # Sample galaxy magnitudes
+    n_galaxies = galaxy_kwargs.get('size')
+    magnitude = schechter_lf_magnitude(**galaxy_kwargs)
```
ntessore, line 81:
```suggestion
    magnitude = schechter_lf_magnitude(**galaxy_kwargs)
    n_galaxies = len(magnitude)
```

skypy/halo/abundance_matching.py
```diff
@@ -14,36 +14,75 @@
...
+    indexing[1:] = np.cumsum(n_subhalos)
+    total_subhalos = indexing[-1]
+    subhalo_group = np.empty(total_subhalos)
+    for first, last, id in zip(indexing[:-1], indexing[1:], halo_group):
+        subhalo_group[first:last] = id
+
+    # Concatenate halos and subhalos
+    mass = np.concatenate([halo_mass, subhalo_mass])
+    group = np.concatenate([halo_group, subhalo_group])
+    parent = np.array((True,) * n_halos + (False,) * total_subhalos)
```
ntessore, line 77:
```suggestion
    parent = np.empty(n_halos+total_subhalos, dtype=bool)
    parent[:n_halos] = True
    parent[n_halos:] = False
```
just because that is creating a potentially huge but unnecessary intermediary tuple

skypy/halo/abundance_matching.py
```diff
@@ -14,36 +14,75 @@
...
+    # Sample galaxy magnitudes
+    n_galaxies = galaxy_kwargs.get('size')
+    magnitude = schechter_lf_magnitude(**galaxy_kwargs)
+
+    # Sort halos and galaxies by mass and magnitude
+    n_matches = min(n_halos + total_subhalos, n_galaxies)
+    sort_mass = np.argsort(mass)[-n_matches:][::-1]
+    sort_magnitudes = np.argsort(magnitude)[:n_matches]
+
+    return mass[sort_mass], group[sort_mass], parent[sort_mass], magnitude[sort_magnitudes]
```
ntessore, line 88: won't this return a view to the original arrays with effectively random memory locations? it might be better for performance to apply the sort order in place if the arrays are potentially huge

skypy/halo/abundance_matching.py
```diff
@@ -14,36 +14,75 @@
...
+    galaxy_kwargs : dict
+        Dictionary of keyword arguments for skypy.galaxy.schechter_lf_magnitude.
 
     Returns
     -------
-    matched_Table : Astropy Table
-        Table of abundance-matched halos and galaxies.
-
+    mass : array_like
+        Array of (sub)halo masses.
```
Lucia-Fonseca, line 44: Do we want to specify units here?

skypy/halo/abundance_matching.py
```diff
@@ -14,36 +14,75 @@
...
+    group : array_like
+        Array of halo group ids.
```
Lucia-Fonseca, line 46:
```suggestion
        Array of halo group identity number.
```
Avoid abbreviations.

skypy/halo/abundance_matching.py
```diff
@@ -14,36 +14,75 @@
...
     ----------
     .. [1] Vale A., Ostriker J. P., 2004, MNRAS, 353, 189
     """
 
-    halos.sort(mass, reverse=True)
-    galaxies.sort(luminosity, reverse=True)
-    return hstack([halos, galaxies], join_type=join_type)
+    # Sample halo and subhalo masses
+    halo_mass = press_schechter(**halo_kwargs)
+    halo_mass[::-1].sort()
```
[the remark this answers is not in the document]
Lucia-Fonseca: Agreed

skypy/halo/abundance_matching.py
Lucia-Fonseca: Agreed

skypy/halo/abundance_matching.py
Lucia-Fonseca: Agreed

Lucia-Fonseca (changes requested):

skypy/halo/abundance_matching.py
ntessore: oh come on

skypy/halo/abundance_matching.py
```diff
@@ -44,6 +46,13 @@ def vale_ostriker(halos, galaxies, mass='mass', luminosity='luminosity',
     .. [1] Vale A., Ostriker J. P., 2004, MNRAS, 353, 189
```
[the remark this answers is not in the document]
rrjbca: I'm going to avoid adding a docstring example here given the vast number of parameters it takes. There is `examples/abundance_matching.yml`

ntessore (approved):
super clear now

Lucia-Fonseca (approved):
Looks good!

PR #262, opened 2020-09-03
mirror examples to repository

ntessore:
This GitHub workflow should update the skypyproject/examples repository every night, mirroring all branches and tags of our library repository.

### DIFF
--- a/.github/workflows/examples.yaml
+++ b/.github/workflows/examples.yaml
@@ -1,0 +1,66 @@
+name: Mirror Examples
+on:
+  workflow_dispatch:
+  schedule:
+    - cron: '0 4 * * *'
+jobs:
+  examples:
+    runs-on: ubuntu-latest
+    steps:
+    - name: Checkout Repository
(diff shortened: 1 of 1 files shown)


ntessore:
I think there is no way for me to test that this works before it is merged?

rrjbca (commented):
Not sure what  `on: workflow_dispatch` is doing here, but rather than mirroring on a schedule wouldn't it make sense to trigger on push or release like you had originally? You can also trigger e.g. only when the examples are changed on master:

```
on:
  push:
    branches:
    - master
    paths:
    - 'examples/**'
```

ntessore:
> Not sure what `on: workflow_dispatch` is doing here

One can manually trigger a `workflow_dispatch` which would be useful if the README on the examples master branch is edited and the changes are to be deployed to all other branches and tags before the next regular trigger.

rrjbca (approved):
Looks good, will merge now

PR #263, opened 2020-09-04
flag to show skypy version

ntessore:
Adds a flag to the `skypy` command that outputs the version of the library.

```bash
$ skypy --version
0.1.dev141+g84a818c
```

### DIFF
--- a/skypy/pipeline/scripts/skypy.py
+++ b/skypy/pipeline/scripts/skypy.py
@@ -35,6 +35,7 @@ values as keyword arguments.
 """
 
 import argparse
+from skypy import __version__ as skypy_version
 from skypy.pipeline.driver import SkyPyDriver
 import sys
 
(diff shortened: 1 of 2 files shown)


rrjbca (commented):
Do we need a test? Or overkill? I would suggest:

```
from contextlib import redirect_stdout
from io import StringIO
from skypy import __version__

# Argparse version
version = StringIO()
with pytest.raises(SystemExit) as e:
    with redirect_stdout(version):
        skypy.main(['--version'])
assert version.getvalue().strip() == __version__
assert e.value.code == 0
```

ntessore:
I thought it unnecessary to test argparse, but perhaps this guards against someone removing the version flag. Feel free to commit here

rrjbca (approved):
Yep, the test also guards against any changes in how argparse handles `action='version'`. Looks good now will merge

PR #273, opened 2020-09-11
weights for Dirichlet coefficients

ntessore:
## Description

Adds a `weight` parameter to `skypy.galaxy.spectra.dirichlet_coefficients` to individually scale each component.

## Checklist
- [x] Write unit tests


rrjbca (commented):
I don't understand why this is needed, perhaps in part because all the examples you give have `scale=1` i.e. nothing is changed from before. What new functionality are we trying to support with this? Is this to allow rescaling by the sampled B-band luminosity or stellar mass? Or the rescaling described in Herbel et al. Appendix C?

ntessore:
> I don't understand why this is needed, perhaps in part because all the examples you give have `scale=1` i.e. nothing is changed from before. What new functionality are we trying to support with this? Is this to allow rescaling by the sampled B-band luminosity or stellar mass? Or the rescaling described in Herbel et al. Appendix C?

If `C` is the 5-tuple of coefficients from a kcorrect template fit, `C` is not Dirichlet-distributed, but `C/scale` is, where `scale` has 5 vastly different entries.

rrjbca:
Right, so because the templates are flux per unit stellar mass, in this case scale would be the stellar mass? Or if you were sampling B-band luminosity scale would be something like L * dM/dL?

ntessore:
No, scale is just a numerical factor. If you look at the marginal distributions of the coefficients from a galaxy sample, they do individually look beta-distributed, but at vastly different scales. After taking that scale out, the coefficients (with whatever units they still have attached) are more or less Dirichlet-distributed.

rrjbca:
OK, so then where *does* the sampled stellar mass / B-band luminosity fit into this pipeline?

philipp128:
I agree. We should include this scaling (previously we called it weighting). In my current analysis, I also assume that this weighting includes the solar mass unit.

ntessore:
> OK, so then where _does_ the sampled stellar mass / B-band luminosity fit into this pipeline?

We will eventually need to have two new functions to normalise the magnitudes:

- Normalise to a reference band: This is when sampling from a Schechter LF, where the sampled magnitudes are the B-band magnitudes. The coefficients and templates will also produce a B-band magnitude, and all magnitudes will be renormalised so that template-B-band matched sampled-B-band.
- Normalise to stellar mass: This is when a stellar mass is given. The coefficients need to be summed, the distance modulus taken out, and the magnitudes will be normalised such that the result (i.e. stellar mass) agrees with the desired value.

philipp128:
> OK, so then where _does_ the sampled stellar mass / B-band luminosity fit into this pipeline?

What exactly do you mean? For sampling the B-band luminosity nothing changed but it's getting easier to include the weights/stellar mass if it is not given.

And if a stellar mass is given, we just need to scale such that the sum of the coefficients corresponds to the mass.

rrjbca (changes requested):
- Can you restore the comment describing the Dirichlet sampling algorithm, I expect it wont be obvious to most users
- Can you also update `examples/mccl_galaxies.yml` with the new scale parameter
- Could you add a test where `scale` is not just a scalar but a 1D array of the same size as the order of the Dirichlet distribution (and perhaps one where it is the wrong size and it fails)

rrjbca (approved):
Looks good. I'm happy to merge as is but offer a couple of thoughts below:
- Is `dirichlet_coefficients` still an accurate name for this function? Since the function does not return Dirichlet samples (they are rescaled) and the purpose is specifically to generate template coefficients, perhaps `template_coefficients` or something similar would be more appropriate?
- Should `scale` be an optional kwarg? Following on from above, if we were implementing a Dirichlet sampler I would say it should be optional, but if we are implementing a model for galaxy template coefficients then scale is an essential part of that model and it should remain as required.
- Maybe update the docstring slightly to reflect the new rescaling e.g. `Galaxy SED template coefficients from a rescaled Dirichlet distribution.`

philipp128:
> Looks good. I'm happy to merge as is but offer a couple of thoughts below:
> 
> * Is `dirichlet_coefficients` still an accurate name for this function? Since the function does not return Dirichlet samples (they are rescaled) and the purpose is specifically to generate template coefficients, perhaps `template_coefficients` or something similar would be more appropriate?
> * Should `scale` be an optional kwarg? Following on from above, if we were implementing a Dirichlet sampler I would say it should be optional, but if we are implementing a model for galaxy template coefficients then scale is an essential part of that model and it should remain as required.
> * Maybe update the docstring slightly to reflect the new rescaling e.g. `Galaxy SED template coefficients from a rescaled Dirichlet distribution.`

I agree with these suggestions.

rrjbca (approved):

PR #271, opened 2020-09-09
utility decorators

ntessore:
## Description

This PR adds a number of useful decorators to `skypy.utils`. See the generated documentation for details:

- [Use the default cosmology](https://1630-221432358-gh.circle-artifacts.com/0/html/utils/index.html#uses-default-cosmology)
- [Broadcast arguments to same shape](https://1630-221432358-gh.circle-artifacts.com/0/html/utils/index.html#broadcast-arguments)
- [Evaluate dependent arguments](https://1630-221432358-gh.circle-artifacts.com/0/html/utils/index.html#dependent-argument)

## Checklist
- [x] Follow the [Contributor Guidelines](https://github.com/skypyproject/skypy/blob/master/CONTRIBUTING.md)
- [x] Write unit tests
- [x] Write documentation strings
- [x] Assign someone from your working team to review this pull request
- [x] Assign someone from the infrastructure team to review this pull request

PR #274, opened 2020-09-11
functions for working with magnitudes

ntessore:
## Description

Adds a couple of functions to `skypy.galaxy.luminosity` that deal with magnitudes:

- `apparent_to_absolute_magnitude`
- `absolute_to_apparent_magnitude`
- `distance_modulus`

These are necessary to easily perform the conversions from config files.

The PR also moves the luminosity conversion functions into the same module:

- `luminosity_from_absolute_magnitude`
- `absolute_magnitude_from_luminosity`

These now optionally take a string argument for `zeropoint` which can be one of the bandmass magnitudes for reference luminosities from the `luminosity_in_band` dictionary.

The functions are currently build on top of #271 but could do without if that PR is not to be merged.

<del>The PR is also built on top of #275 so the other PR should be merged first.</del>

## Checklist
- [x] Write unit tests


skypy/galaxy/luminosity.py
```diff
@@ -4,15 +4,158 @@
...
+    'apparent_to_absolute_magnitude',
+    'distance_modulus',
+    'luminosity_in_band',
+    'luminosity_from_absolute_magnitude',
+    'absolute_magnitude_from_luminosity',
     'herbel_luminosities',
 ]
 
 
+def absolute_to_apparent_magnitude(M, DM):
```
rrjbca, line 22:
```suggestion
def absolute_to_apparent_magnitude(absolute_magnitude, distance_modulus=None, redshift=None, cosmology=None):
    if not distance_modulus:
        distance_modulus = skypy.distance_modulus(redshift, cosmology)
```
I think we need to support this conversion given a redshift so that pipeline users don't have to calculate the distance modulus as an intermediate column. Could use this suggestion or implement an additional function

rrjbca: Also use full names for variables rather than abbreviations or symbolic representations where practical

rrjbca:
Would it make sense for `skypy/galaxy/luminosity.py` to only contain luminosity samplers and move most of the new functions in this PR to a new `skypy/photometry/conversions.py` module? They are certainly general enough to use for stars, supernovae...

ntessore:
> Would it make sense for `skypy/galaxy/luminosity.py` to only contain luminosity samplers and move most of the new functions in this PR to a new `skypy/photometry/conversions.py` module? They are certainly general enough to use for stars, supernovae...

Yes, everything from `galaxy.spectrum` and these from `galaxy.luminosity` should ideally go into `photometry`, no need for submodules. But that is a separate changeset.

skypy/galaxy/luminosity.py
ntessore: My reasoning was that one would always want to keep the distance modulus around after computation, because it takes a while to compute, and I don't like either-or function signatures. If a user wants to not keep the distance modulus around and compute it in one go, they can chain the function call.

ntessore: And `absolute_magnitude` and `distance_modulus` are way too long words to force anyone to type them out...

philipp128: But later in the use case, the user don't really have to type them out.

skypy/galaxy/luminosity.py
```diff
@@ -9,10 +9,153 @@
...
+    'Lsun_B': 5.31,
+    'Lsun_V': 4.80,
+    'Lsun_R': 4.60,
+    'Lsun_I': 4.51,
+}
+'''Bandpass magnitude of reference luminosities.
+
+These values can be used for conversion in `absolute_magnitude_from_luminosity`
+and `luminosity_from_absolute_magnitude`. The `Lsun_{UBVRI}` values contain the
+absolute magnitude of the sun in Johnson/Cousins bands from [1]_.
```
rrjbca, line 94:
```suggestion
AB absolute magnitude of the sun in Johnson/Cousins bands from [1]_.
```
I assume these are AB magnitudes? In which case state that specifically in the documentation. Otherwise the rest of this PR looks fine.

rrjbca:
I also notice that the old `luminosity_from_absolute_magnitude` function is used in both the abundance matching and gravitational wave examples so could you update those config file with the new module path as well please.

ntessore:
> I also notice that the old `luminosity_from_absolute_magnitude` function is used in both the abundance matching and gravitational wave examples so could you update those config file with the new module path as well please.

Done. At some point we should go through all of these naked floats and turn them into quantities (here e.g. with `mag(AB)` units) so that all of these distinctions are made explicit.

rrjbca (approved):

rrjbca (commented):
Do the examples also need to specify the reference band zeropoint? i.e:
```
luminosity: !skypy.galaxy.luminosity.luminosity_from_absolute_magnitude
  absolute_magnitude: $merger_rates.magnitude
  zeropoint: Lsun_B
```

examples/abundance_matching.yml
```diff
@@ -30,7 +30,7 @@ tables:
       alpha: -0.5
       m_lim: 35
       size: 100
-    luminosity: !skypy.utils.astronomy.luminosity_from_absolute_magnitude
+    luminosity: !skypy.galaxy.luminosity.luminosity_from_absolute_magnitude
       absolute_magnitude: $galaxies.magnitude
```
rrjbca, line 34:
```suggestion
    luminosity: !skypy.galaxy.luminosity.luminosity_from_absolute_magnitude
      absolute_magnitude: $galaxies.magnitude
      zeropoint: Lsun_B
```

rrjbca (approved):

PR #275, opened 2020-09-11
change mass quenching p-values

ntessore:
## Description

The mass quenching tests sometimes fail. This PR changes the p-value from 0.05 to 0.01, which is the number most tests use.

[This PR fixes the issue "the test fails". Perhaps one should also investigate whether everything is in order with the test.]

## Checklist
- [x] Write unit tests


rrjbca (approved):
Approving this for consistency with other KS tests in the codebase. But at some point it is probably still worth checking that *all* such tests are valid and have been implemented properly; these are not the only ones to have failed recently.

PR #276, opened 2020-09-11
AB magnitudes from composites of template sepctra

rrjbca:
## Description
Calculate the AB magnitudes in a set of given bandpasess for object whose spectra are composites of a set of template spectra. Optionally redshift the spectra. For a scalar redshift or for a small number of objects, the magnitudes are computed exactly. For a large number of objects each at different redshifts, the template magnitudes are calculated on a fixed grid of redshifts and then interpolated at the redshifts of each object.

## Checklist
- [x] Write unit tests


ntessore:
Couple of suggestions to take or leave:

- I think this should be `magnitude_from_templates` because the coefficients determine if absolute or apparent, depending on whether they have the distance modulus factored in (and since they are generally empirical, they generally do).
- Merging PR #277 would also simplify the interface of this function.
- Going even further, this function could ideally take a `specutils.SpectrumCollection`, check that the axes of the coefficients matches the size of the collection, and have everything neatly packaged. I have a custom `kcorrect` loader for specutils that can do the loading.

philipp128 (commented):
It looks very good! However, I agree with @ntessore that the name should be magnitude_from_templates or sth like this more general.

philipp128:
> * Going even further, this function could ideally take a ~`specutils.SpectrumCollection`~`specutils.SpectrumList`, check that the axes of the coefficients matches the length of the list, and have everything neatly packaged. I have a custom `kcorrect` loader for specutils that can do the loading.

If I understand it correctly the `spectulis.SpectrumList` object is just a list of spectra? In this case, I think it's a nice idea to enable this as an input.

ntessore:
> > * Going even further, this function could ideally take a ~`specutils.SpectrumCollection`~`specutils.SpectrumList`, check that the axes of the coefficients matches the length of the list, and have everything neatly packaged. I have a custom `kcorrect` loader for specutils that can do the loading.
> 
> If I understand it correctly the `spectulis.SpectrumList` object is just a list of spectra? In this case, I think it's a nice idea to enable this as an input.

It is a list of spectra, but `SpectrumList` has a `read()` method and allows definition of custom loader formats. So one can define the `skypy spectrum templates` format, call `SpectrumList.read('kcorrect', format='skypy spectrum templates')`, and the loader will resolve all of this automatically (which works really well).

skypy/galaxy/spectrum.py
```diff
@@ -197,6 +198,77 @@ def mag_ab(spectrum, bandpass, redshift=None):
     return mag_ab.item() if not return_shape else mag_ab.reshape(return_shape)
 
 
+def magnitudes_from_templates(coefficients, templates, bandpasses, redshift=None,
+                              resolution=1000, stellar_mass=None, distance_modulus=None):
```
ntessore, line 202: Could this be `bandpass` just for consistency with `mag_ab`? I know this is quite an annoying request for all the follow-up changes below so feel free to disregard.

skypy/galaxy/spectrum.py
```diff
@@ -197,6 +198,77 @@ def mag_ab(spectrum, bandpass, redshift=None):
     return mag_ab.item() if not return_shape else mag_ab.reshape(return_shape)
```
ntessore, line 200: This function could be decorated with `@spectral_data_input` so that keywords like `templates='kcorrect_spec', bandpass='DECam_griz'` would be accepted.

ntessore (approved):
I approve this with two suggestions, I think it's fine to merge as it is and tackle them separately.

ntessore (approved):

PR #280, opened 2020-09-16
fix schechter random sampling test failures

ntessore:
## Description

The tests for `skypy.utils.random.schechter` can fail randomly when the shape parameter `alpha` makes the distribution so steep that the `resolution` is not enough to produce a good sample. This moves the range of `alpha` to where it is less steep so that this should not happen.

## Checklist
- [ ] Write unit tests

### DIFF
--- a/skypy/utils/tests/test_random.py
+++ b/skypy/utils/tests/test_random.py
@@ -43,8 +43,8 @@ def test_schechter_gamma():
 
     from skypy.utils.random import schechter
 
-    # when alpha > -1, x_min ≈ 0, x_max ≈ ∞, distribution is gamma
-    alpha = np.random.uniform(-1, 2)
+    # when alpha > 0, x_min ≈ 0, x_max ≈ ∞, distribution is gamma
+    alpha = np.random.uniform(0, 2)
     x_min = 1e-20
     x_max = 1e+20
     scale = 2.5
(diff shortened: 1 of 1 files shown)


rrjbca (commented):
Is this preferable to increasing the resolution used in the test?
```
x = schechter(alpha, x_min, x_max, size=1000, resolution=10000)
```

ntessore:
> Is this preferable to increasing the resolution used in the test?
> 
> ```
> x = schechter(alpha, x_min, x_max, size=1000, resolution=10000)
> ```

The resolution in the test is already absurdly high, but if alpha < 0 is allowed you can get it steep enough to make the test fail.

skypy/utils/tests/test_random.py
```diff
@@ -43,8 +43,8 @@ def test_schechter_gamma():
 
     from skypy.utils.random import schechter
 
-    # when alpha > -1, x_min ≈ 0, x_max ≈ ∞, distribution is gamma
-    alpha = np.random.uniform(-1, 2)
+    # when alpha > 0, x_min ≈ 0, x_max ≈ ∞, distribution is gamma
```
rrjbca, line 46:
```suggestion
    # when alpha > 0, x_min ≈ 0, x_max ≈ ∞, distribution is gamma
    # n.b. if alpha < 0 the distribution becomes too steep to resolve accurately
```

rrjbca (approved):

PR #284, opened 2020-09-22
Halo circular velocity function from Maller & Bullock 2004

sarahbridle:
## Checklist
- [ ] Write unit tests


rrjbca (changes requested):
Great start! A few suggestions to improve the code:
- We prefer to avoid abbreviations in variable names where practical, so `M` should probably be renamed `mass`
- We try to use astropy units where appropriate. So you should do `from astropy.units import Unit` and then e.g. `96.6 * Unit('km s-1')` and `(mass.to_value('Msun') / 1e11)`
- Use `cosmology.h` rather than `cosmology.H0 / 100`. This also has the correct units.
- We need a test for the function. See e.g. [test_carroll](https://github.com/skypyproject/skypy/blob/master/skypy/power_spectrum/tests/test_growth.py) for a simple example of writing tests. Create a new file `skypy/halo/tests/test_properties.py` and write a function `test_circular_velocity`. An easy test would be to choose the arguments such that each bracket in the function equals one and check that the returned value equals 96.6 km/s. Perhaps you can think of others
- There are some trivial formatting issues relating to whitespace/tabs/blank lines etc. At the bottom of the pull request next to "codeclimate" click "details" to see the suggestions

skypy/halo/properties.py
```diff
@@ -0,0 +1,76 @@
...
+    >>> redshift = np.arange(0.3, 1, 0.5)
+    >>> properties.halo_circular_velocity(M, Delta_v, redshift, cosmology)
+    <Quantity [ 6.11303684, 36.72661831] km2 / (Mpc2 s2)>
+
+    References
+    ----------
+    .. [1] Maller and Bullock 2004 MNRAS 355 694 DOI:10.1111/j.1365-2966.2004.08349.x
+    
+	Notes of things to ask / think about:
+	* where to get Delta_v from - do we have a module for this already, or compute it in here too?
```
Lucia-Fonseca, line 66: we do not have a module for this yet. We could write its own function if we would reuse it or leave it as a free parameter.

skypy/halo/properties.py
```diff
@@ -0,0 +1,76 @@
...
+	* fix h hack / check h in Maller & Bullock is h100
+	* should we call it the circular_velocity or the virial_velocity?
```
Lucia-Fonseca, line 68: I would call it as the authors do, virial velocity, to avoid confusion.

skypy/halo/properties.py
```diff
@@ -0,0 +1,76 @@
...
+import numpy as np
+
+__all__ = [
+    'halo_circular_velocity',
+ ]
+
+def halo_circular_velocity(M, Delta_v, redshift, cosmology):
+    """Halo circular velocity.
+    This function computes the halo circular velocity, setting it 
+    equal to the virial velocity using equation (3) from [1]_.
```
Lucia-Fonseca, line 25:
```suggestion
    equal to the virial velocity using equation (3) and footnote 2 from [1]_.
```

skypy/halo/properties.py
```diff
@@ -0,0 +1,76 @@
...
+
+    References
+    ----------
+    .. [1] Maller and Bullock 2004 MNRAS 355 694 DOI:10.1111/j.1365-2966.2004.08349.x
+    
+	Notes of things to ask / think about:
+	* where to get Delta_v from - do we have a module for this already, or compute it in here too?
+	* fix h hack / check h in Maller & Bullock is h100
+	* should we call it the circular_velocity or the virial_velocity?
+	* ditto M - should we call it M_v or pretend virial mass = halo mass throughout?
```
Lucia-Fonseca, line 69: My preference would be `M_v`, as in the paper.

skypy/halo/properties.py
```diff
@@ -0,0 +1,76 @@
...
+	* should we output the virial radius here as well or is this already done elsewhere?
+    """
+
+    h = cosmology.H0 / 100
+    circular_velocity = 96.6 * (Delta_v * cosmology.Om0 * h **2) * pow((1 + redshift)/0.3,0.5) * pow(M/1e11,1/3)
```
Lucia-Fonseca, line 74: Does this correspond to the equation in footnote 2?

Lucia-Fonseca (changes requested):
It needs some changes and tests.
Please, take this review as suggestions.

Lucia-Fonseca:
We should update the description to link with the correspondent issue.

skypy/halo/properties.py
```diff
@@ -0,0 +1,76 @@
...
+    >>> cosmology = Planck15
+    >>> M = 10**np.arange(9.0, 12.0, 2)
+    >>> Delta_v = np.arange(1.0, 1.1, 0.1)
+    >>> redshift = np.arange(0.3, 1, 0.5)
+    >>> properties.halo_circular_velocity(M, Delta_v, redshift, cosmology)
+    <Quantity [ 6.11303684, 36.72661831] km2 / (Mpc2 s2)>
+
+    References
+    ----------
+    .. [1] Maller and Bullock 2004 MNRAS 355 694 DOI:10.1111/j.1365-2966.2004.08349.x
```
Lucia-Fonseca, line 63: I believe the equation comes from this other paper
https://arxiv.org/pdf/1403.1873.pdf

skypy/halo/properties.py
Lucia-Fonseca, line 74: Perhaps I am confusing the equation but I believe it should be
```suggestion
    circular_velocity = 96.6 * np.power(Delta_v * cosmology.Om0 * h **2 / 24.4, 1.0 / 6.0) * np.sqrt((1 + redshift) / 3.3) * np.power(M / 1.0e11 , 1.0 / 3.0)
```

skypy/halo/properties.py
sarahbridle: Great, thanks - lets make a separate function so we can call that independently if we want to (and sub in different versions of the Delta_v code)

skypy/halo/properties.py
sarahbridle: Great, thanks. We discussed to go with halo_virial_mass !

skypy/halo/properties.py
```diff
@@ -0,0 +1,75 @@
...
+    ----------
+    M : (nm,) array_like
+    	Array for the virial mass, in units of solar mass.
+	Delta_v : (nm,) array_like
+		The mean overdensity of the halo.
+	redshift : (nm,) array_like
+		The redshift of each halo.
+	cosmology : astropy.cosmology.Cosmology
+        Cosmology object providing methods for the evolution history of
+        omega_matter and omega_lambda with redshift.
```
Lucia-Fonseca, line 37: Over indented block

skypy/halo/properties.py
```diff
@@ -0,0 +1,76 @@
...
+    References
+    ----------
+    .. [1] Maller and Bullock 2004 MNRAS 355 694 DOI:10.1111/j.1365-2966.2004.08349.x
+    
+	Notes of things to ask / think about:
+	* where to get Delta_v from - do we have a module for this already, or compute it in here too?
+	* fix h hack / check h in Maller & Bullock is h100
+	* should we call it the circular_velocity or the virial_velocity?
+	* ditto M - should we call it M_v or pretend virial mass = halo mass throughout?
+	* should we output the virial radius here as well or is this already done elsewhere?
```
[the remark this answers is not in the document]
sarahbridle: Good plan. Will make a new function, thanks.

skypy/halo/properties.py
sarahbridle: will do!

skypy/halo/properties.py
sarahbridle: hopefully now fixed thanks!

skypy/halo/properties.py
sarahbridle: Thanks! I need to look into this carefully!

sarahbridle: thanks, you're right - fixed!

skypy/halo/properties.py
sarahbridle: argggh, you're totally right, thank you!

itrharrison:
Replaced with #529 to change target branch.

PR #291, opened 2020-09-26
improved loading of multiple spectral data

ntessore:
## Description

Make `load_spectral_data` behave as expected when multiple compatible spectra are requested:

- `load_spectral_data('DECam_g')` returns a `(1441,)` flux array
- `load_spectral_data('DECam_gr')` returns a `(2, 1441)` flux array
- `load_spectral_data(['DECam_gr', 'DECam_i'])` returns a `(3, 1441)` flux array equal to `load_spectral_data('DECam_gri')`
- `load_spectral_data(['DECam_gr', 'DECam_iz'])` returns a `(4, 1441)` flux array equal to `load_spectral_data('DECam_griz')`

## Checklist
- [x] Write unit tests


rrjbca:
> * `load_spectral_data(['DECam_gr', 'DECam_iz'])` returns a `(2, 2, 1441)` flux array

Is there a use case where this behaviour is applicable? I can't think that we would ever want 3D spectral quantities, particularly as everything is designed to end up in table columns. It would make much more sense to me if this yielded an array with shape `(4, 1441)`

ntessore:
> Is there a use case where this behaviour is applicable? I can't think that we would ever want 3D spectral quantities, particularly as everything is designed to end up in table columns. It would make much more sense to me if this yielded an array with shape `(4, 1441)`

Tricky one. I am coming at this from using `load_spectral_data(['kcorrect_spec', 'kcorrect_rawspec'])` to test multiple sets of templates, which should be in shape `(2, 5, n)`, but you are right that for the the filters it's better to stack the arrays if possible. Multiple filters is probably the more common occurrence?

rrjbca:
> Tricky one. I am coming at this from using `load_spectral_data(['kcorrect_spec', 'kcorrect_rawspec'])` to test multiple sets of templates, which should be in shape `(2, 5, n)`, but you are right that for the the filters it's better to stack the arrays if possible. Multiple filters is probably the more common occurrence?

I think loading and storing each template set separately makes sense. We don't currently have any functions that (are intended to) work on multiple template sets that would require loading and storing them together in a single object. In general it makes sense to me that a user would want multiband photometry but doesn't make sense that they would want multiple different model spectra for the same galaxies. If a user did want to get magnitudes for both `kcorrect_spec` and `kcorrect_rawspec` would it not be reasonable to expect them to load them separately?

i.e. `magnitudes_from_templates(..., "kcorrect_spec", ["DECam_r", "Johnson_B"], ...)` makes sense to me but `magnitudes_from_templates(..., ["kcorrect_spec", "kcorrect_rawspec"], "DECam_r", ...)` doesn't.

rrjbca (approved):

PR #301, opened 2020-09-30
Use default cosmology context in pipeline

### LINKED ISSUE #293: set default cosmology from pipeline
When a `cosmology` is set in a pipeline configuration, it should be registered as the default cosmology during the run:

```py
with default_cosmology.set(...):
    ...
```

This probably needs taking the `cosmology` variable out of the normal dependency graph and process it before anything else is run.

rrjbca:
## Description
Pipeline now sets `parameters` and `cosmology` before executing all other jobs in the DAG. The DAG is also resolved using the `default_cosmology.set()` context so that all functions decorated with `@uses_default_cosmology` will use the cosmology defined in the config (or the astropy default cosmology if undefined). Note that currently `cosmology` can only depend on values in parameters and not any other variables defined in the config file. Resolves #293 

## Checklist
- [x] Write unit tests

### DIFF
--- a/skypy/pipeline/_pipeline.py
+++ b/skypy/pipeline/_pipeline.py
@@ -4,8 +4,9 @@ This module provides methods to run pipelines of functions with dependencies
 and handle their results.
 """
 
+from astropy.cosmology import default_cosmology
 from astropy.table import Table
-from copy import deepcopy
+from copy import copy, deepcopy
 from importlib import import_module
 import builtins
 import networkx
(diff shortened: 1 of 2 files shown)


skypy/pipeline/_pipeline.py
```diff
@@ -126,8 +128,6 @@ def __init__(self, configuration):
         # - keep track where functions need to be called
         #   functions are tuples (function name, [function args])
         functions = {}
-        for job in self.parameters:
-            self.dag.add_node(job)
```
ntessore, line 130: We do need the nodes in the DAG though, or the check in line 164 will fail for anything that depends on a parameter. Same with the `cosmology` node.

ntessore (changes requested):
Looks great! I think the DAG would complain if anything depends explicitly on a parameter value at the moment. This would cause the [failure in the docs](https://skypy--301.org.readthedocs.build/en/301/pipeline/index.html) where the plot does not show up.

ntessore (approved):
🎉

PR #344, opened 2020-10-09
sample positions uniformly over HEALPix pixel

ntessore:
## Description

Adds the `uniform_in_pixel` function to uniformly sample positions over an individual HEALPix pixel.

## Checklist
- [x] Write unit tests

### DIFF
--- a/skypy/position/_uniform.py
+++ b/skypy/position/_uniform.py
@@ -48,7 +48,7 @@ def uniform_around(centre, area, size):
 
 
 def uniform_in_pixel(nside, ipix, size, nest=False):
-    from healpy.pixelfunc import max_pixrad, pix2ang, ang2pix
+    from healpy import max_pixrad, pix2ang, ang2pix
 
     # get the centre of the healpix pixel as a SkyCoord
     centre_lon, centre_lat = pix2ang(nside, ipix, nest=False, lonlat=True)
--- a/skypy/position/_uniform.py
+++ b/skypy/position/_uniform.py
@@ -2,9 +2,9 @@
 
 import numpy as np
 from astropy import units
+from astropy.coordinates import SkyCoord
 
 TWO_PI = 2*np.pi
-PI_HALF = np.pi/2
 
 
 @units.quantity_input(area=units.sr)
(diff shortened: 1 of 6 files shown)


skypy/position/tests/test_healpy.py
```diff
@@ -0,0 +1,46 @@
...
+    nside = 2
+    npix = healpy.nside2npix(nside)
+
+    # sample entire sky with 20 positions per pixel
+    theta, phi = np.empty(npix*20), np.empty(npix*20)
+    for ipix in range(npix):
+        pos = uniform_in_pixel(nside, ipix, size=20)
+        assert len(pos) == 20
+        theta[ipix*20:(ipix+1)*20] = np.pi/2 - pos.dec.rad
+        phi[ipix*20:(ipix+1)*20] = pos.ra.rad
```
rrjbca, line 40:
```suggestion
    size = 20
    theta, phi = np.empty(npix*size), np.empty(npix*size)
    for ipix in range(npix):
        pos = uniform_in_pixel(nside, ipix, size=size)
        assert len(pos) == size
        theta[ipix*size:(ipix+1)*size] = np.pi/2 - pos.dec.rad
        phi[ipix*size:(ipix+1)*size] = pos.ra.rad
```

skypy/position/tests/test_healpy.py
```diff
@@ -0,0 +1,46 @@
...
+    from skypy.position import uniform_in_pixel
+
+    # 768 healpix pixels
+    nside = 8
+    npix = healpy.nside2npix(nside)
+
+    ipix = np.random.randint(npix)
+
+    pos = uniform_in_pixel(nside, ipix, size=1000)
+    assert len(pos) == 1000
```
rrjbca, line 17:
```suggestion
    size = 1000
    pos = uniform_in_pixel(nside, ipix, size=size)
    assert len(pos) == size
```

skypy/position/_uniform.py
```diff
@@ -45,3 +45,75 @@ def uniform_around(centre, area, size):
 
     # construct random sky coordinates around centre
     return centre.directional_offset_by(phi, theta)
+
+
+def uniform_in_pixel(nside, ipix, size, nest=False):
+    '''Uniform distribution of points over healpix pixel.
+
+    Draws randomly distributed points from the healpix pixel `ipix` for a map
+    of with `nside` parameter.
```
rrjbca, line 54:
```suggestion
    with a given `nside` parameter.
```

skypy/position/_uniform.py
```diff
@@ -45,3 +45,75 @@ def uniform_around(centre, area, size):
...
+    ----------
+    nside : int
+        Healpix map `nside` parameter.
+    ipix : int
+        Healpix map pixel index.
+    size : int
+        Number of points to draw.
+    nest : bool, optional
+        Assume ``NESTED`` pixel ordering, or ``RING`` pixel ordering otherwise.
+        Default is ``RING`` pixel ordering.
```
rrjbca, line 66:
```suggestion
        If True assume ``NESTED`` pixel ordering, otherwise ``RING`` pixel
        ordering. Default is ``RING`` pixel ordering.
```

rrjbca (changes requested):
One concern I have is there is no protection against an implementation error resulting in `over` being less than one, which could lead to getting stuck in the while loop. Perhaps defining a separate function `_oversampling_factor(nside)` we could test that the result is always greater than one for different values of `nside`? Thoughts?

ntessore:
> One concern I have is there is no protection against an implementation error resulting in `over` being less than one

Over is the area of a circle covering a polygon, divided by the area of a polygon. It cannot be less than one, unless, as you say, the healp* implementation is wrong, in which case we cannot rely on anything else working either. So I don't see why we would worry about it.

skypy/position/_uniform.py
```diff
@@ -45,3 +45,75 @@ def uniform_around(centre, area, size):
...
+    over = area.value/nside2pixarea(nside)
+
+    # the array of longitudes and latitudes of the sample
+    lon, lat = np.empty(0), np.empty(0)
+
+    # rejection sampling over irregularly shaped healpix pixels
+    miss = size
+    while miss > 0:
+        # get the coordinates in a circular aperture around centre
+        sample = uniform_around(centre, area, int(miss*over))
```
ntessore, line 105:
```suggestion
        sample = uniform_around(centre, area, int(np.ceil(miss*over)))
```

rrjbca (approved):

PR #347, opened 2020-10-09
added P&H2017 neutral hydrogen mass model

### LINKED ISSUE #234: Padmanabhan & Refregier HI halo model
## Description

The model of https://arxiv.org/abs/1607.01021 generates a neutral Hydrogen (HI) mass for a halo, given its halo mass and a model for circular velocity. Being able to predict HI masses is a crucial component for us to be able to generate HI observables like 21cm Intensity Maps.

## Inputs
 - Halo mass
- Halo circular velocity (model)

## Outputs
 - Halo neutral hydrogen mass
 - Halo neutral hydrogen radial density profile

## References
 - https://arxiv.org/abs/1607.01021
 - https://arxiv.org/abs/1611.06235
 - https://arxiv.org/abs/1909.11104

itrharrison:
## Description

Merging will close #234. This implements the Padmanabhan & Refregier HI halo mass model described in that issue.

## Checklist
- [X] Write unit tests

### DIFF
--- a/skypy/neutral_hydrogen/__init__.py
+++ b/skypy/neutral_hydrogen/__init__.py
@@ -1,0 +1,16 @@
+"""
+This module contains methods that model the properties of neutral hydrogen
+emission in the Universe.
+
+Merger Rates
+================
+
+.. autosummary::
+   :nosignatures:
+   :toctree: ../api/
+
+   pr_halo_model
+
+"""
+
+from .mass import *  # noqa F401,F403
--- a/skypy/neutral_hydrogen/mass.py
+++ b/skypy/neutral_hydrogen/mass.py
@@ -1,0 +1,85 @@
+import numpy as np
+from astropy import units
+
+
+r"""HI mass module.
+This module provides functions to calculate neutral Hydrogen masses for haloes.
+
+"""
+
+__all__ = [
+           'pr_halo_model',
+]
+
+
+def pr_halo_model(halo_mass, circular_velocity, cosmology,
+                  alpha=0.17,
+                  beta=-0.55,
+                  v_c0=np.exp(1.57) * (units.km / units.s),
+                  v_c1=np.exp(4.39) * (units.km / units.s),
+                  Y_p=0.24):
+    r"""Model of Padmanabhan & Refregier (2017), equation 1.
+
+    Halo neutral Hydrogen (HI 21-cm) masses as a function of halo mass,
+    halo circular velocity, and cosmology.
+
+    Parameters
+    ----------
+    halo_mass : (nhalos,) `~astropy.Quantity`
+        The masses of the underlying halos in units of solar mass.
+    circular_velocity : (nhalos,) `~astropy.Quantity`
+        The circular velocity for the halos in units of [km s-1].
+    cosmology : `~astropy.cosmology.Cosmology`
+        Cosmology object providing values of omega_baryon and omega_matter.
+    alpha : float
+        Linear model constant, fit from data.
+    beta : float
+        Model exponent, fit from data.
+    v_c0 : `~astropy.Quantity`
+        Velocity for the lower exponential cutoff of the model
+        in units of [km s-1].
+    v_c1 : `~astropy.Quantity`
+        Velocity for the upper exponential cutoff of the model
+        in units of [km s-1].
+    Y_p : float
+        Cosmic Helium fraction.
+
+    Returns
+    -------
+    m_hone : (nhalos,) `~astropy.Quantity`
+        Neutral hydrogen mass contained in the halo, in units of solar mass.
+
+    References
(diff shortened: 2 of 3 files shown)


rrjbca (commented):
Since this is all built on halo properties I would put it in `skypy/halo/gas.py` (or similar) rather than starting an entire new module

itrharrison:
@rrjbca hmm... I'm not sure I agree. HI would be an observable which is fundamentally and conceptually different to the other types of observables we have (galaxies, gravy waves etc) and on a level with them.

We could say that 'halos' implies 'Dark Matter halos'. This model also makes a prediction only for HI mass.

By your argument, could we also, say, not suggest that the gravitational wave merger rates should also have been placed in galaxies?

I can imagine there are things where what you say does make sense though. e.g. multi-wavelength observations of a cluster's gas (HI, X-ray) might require some 'gas' structure within the halos module.

laurawolz:
I have had similar thoughts, as the HI models I work with a based on galaxies, so. for this case then, HI model could equally fit into the galaxies module. 

I think, based on the fact that HI can based on halo and galaxy properties (or possibly other observables), it would make most sense to have it as an own module which can host a variety of HI models.

rrjbca:
I meant to say that `neutral_hydrogen` is way too narrow to justify creating an entire new module (in my opinion). If this doesn't belong in halos (probably true) I would at least suggest `skypy/gas/hydrogen.py`. A new module for `gas` has a lot more scope to grow and incorporate e.g. ISM, IGM, ICL etc while `hydrogen.py` can contain *all* HI properties you care about irrespective of wether it is bound to `halos` or `galaxies` and without having to create a new file for each property.

rrjbca:
Following a discussion we decided that `skypy.neutral_hydrogen` is a suitable module for this work. This can always be revisited and changed if we find it conflicts with future design choices.

skypy/neutral_hydrogen/__init__.py
```diff
@@ -1,4 +1,16 @@
-'''This module contains methods that model the properties of neutral hydrogen.
-'''
+"""
+This module contains methods that model the properties of neutral hydrogen
+emission in the Universe.
 
+Merger Rates
```
ntessore, line 5: this is still the heading from GW, plus too many underlines

skypy/neutral_hydrogen/__init__.py
```diff
@@ -1,4 +1,16 @@
...
+================
 
+.. autosummary::
+   :nosignatures:
+   :toctree: ../api/
+
+   pr_halo_model
```
ntessore, line 12: I don't think this is the best name for the model, mostly because it makes me think that this is a halo model. Suggestions would be `hi_mass_pr17` or `padmanabhan_refregier_17` (which not everyone likes).

skypy/neutral_hydrogen/__init__.py
```diff
@@ -1,4 +1,16 @@
...
+
+"""
+
+from .mass import *  # noqa F401,F403
```
ntessore, line 16: It's ambiguous how the module is to be structured. In this way, you could access the function under both `skypy.neutral_hydrogen` as well as `skypy.neutral_hydrogen.mass`. If it's supposed to be the former, then I would add an underscore to `_mass` to indicate it is an implementation file. If the latter, then you should only `from . import mass`. (I guess it depends on how much stuff there is going to be in the module to decide which is right.)

skypy/neutral_hydrogen/mass.py
```diff
@@ -0,0 +1,85 @@
...
+    halo circular velocity, and cosmology.
+
+    Parameters
+    ----------
+    halo_mass : (nhalos,) `~astropy.Quantity`
+        The masses of the underlying halos in units of solar mass.
+    circular_velocity : (nhalos,) `~astropy.Quantity`
+        The circular velocity for the halos in units of [km s-1].
+    cosmology : `~astropy.cosmology.Cosmology`
+        Cosmology object providing values of omega_baryon and omega_matter.
```
ntessore, line 33: This would be better moved to the end of the parameter list because it will be filled by the pipeline runner automatically.

skypy/neutral_hydrogen/tests/test_mass.py
```diff
@@ -0,0 +1,29 @@
+import numpy as np
+import pytest
+
+from astropy.cosmology import FlatLambdaCDM
+from astropy import units
+
+pr_data = np.asarray([0.00000000e+00, 8.08547768e-45, 6.40736312e+09, 3.66714636e+09,
+                      0.00000000e+00, 0.00000000e+00, 0.00000000e+00, 0.00000000e+00]) * units.Msun
```
ntessore, line 8: Is it useful to have half of the tests return zero masses?

Issue #313, opened 2020-10-02
pipeline cannot multi-assign from array


ntessore:
**Describe the bug**

The return value of a function cannot be assigned to multiple columns if it is of array type.

**To Reproduce**

```py
import numpy as np
from skypy.pipeline import Pipeline


def three_columns_as_array():
    return np.random.rand(3, 100)


def three_columns_as_tuples():
    a, b, c = three_columns_as_array()
    return a, b, c


def test_multi_assign():
    config = {
        'tables': {
            'mytable': {
                'col_a, col_b, col_c': (three_columns_as_tuples,)}}}

    pipeline = Pipeline(config)
    pipeline.execute()

    assert pipeline['mytable'].colnames == ['col_a', 'col_b', 'col_c']
    assert len(pipeline['mytable']) == 100

    config['tables']['mytable']['col_a, col_b, col_c'] = (three_columns_as_array,)

    pipeline = Pipeline(config)
    pipeline.execute()

    assert pipeline['mytable'].colnames == ['col_a', 'col_b', 'col_c']
    assert len(pipeline['mytable']) == 100
```

Failure:

```
        config['tables']['mytable']['col_a, col_b, col_c'] = (three_columns_as_array,)

        pipeline = Pipeline(config)
>       pipeline.execute()
...
E           ValueError: Arguments "names" and "dtype" must match number of columns
```

**Expected behavior**
Multi-assignment should work if the output has the correct leading dimension, just like tuple assignment.

ntessore:
Oh I see! Assignment from an array uses the _last_ dimension while assignment from a tuple uses the _first_ dimension. Not sure that this is very intuitive but at least it's not a bug.

PR #349, opened 2020-10-19
Multicolumn assignment call pattern depends on astropy version

rrjbca:
## Description
On the astropy master branch, `Table.add_columns` now expects columns to be passed as a `TableColumns` object and no longer accepts a list of `Column` objects. This breaks our current implementation of multi-column assignment and causes our `dev` tests in the [`Compatibility Workflow`](https://github.com/skypyproject/skypy/actions/runs/313272229) to fail. This PR checks the astropy version to determine which pattern to use when calling `add_columns`. Tested locally using astropy versions 3.2.3, 4.0.2 and master.

## Checklist
- [ ] Write unit tests

### DIFF
--- a/skypy/pipeline/_pipeline.py
+++ b/skypy/pipeline/_pipeline.py
@@ -4,6 +4,7 @@ This module provides methods to run pipelines of functions with dependencies
 and handle their results.
 """
 
+from astropy import __version__ as astropy_version
 from astropy.cosmology import default_cosmology
 from astropy.table import Table
 from copy import copy, deepcopy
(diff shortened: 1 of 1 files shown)


ntessore:
I would still much prefer the assignment to work like tuple assignment in all cases, over the first axis of the right-hand side (cf. #313). Then it becomes a simple loop over pairs of columns on the left, values on the right.

rrjbca:
Although I agree with your suggestion I think it's a separate issue; at the very least before implementing it we need to make a decision on where/how we are going to transpose all of the arrays in our library and also any other functions out in the wild that people want to use in their pipelines. Although this PR may get quickly overwritten by #350 I think we should still merge it just to keep our tests passing the in meantime.

rrjbca:
These changes were merged as part of #353

PR #356, opened 2020-10-29
Pipeline.__getitem__ returns column data as a Quantity or ndarray

### LINKED ISSUE #355: pipeline passes Column objects to functions
**Describe the bug**
The `skypy.pipeline.Pipeline` class passes data as `Column` type to functions. This is problematic when said functions expect a `Quantity`, since `Column` is not entirely `Quantity`-like.

**To Reproduce**
```py
import numpy as np
from astropy import units
from skypy.pipeline import Pipeline


def lengths_in_cm(q):
    return q.to_value(units.cm)


def random_lengths(size):
    return np.random.rand(size)*units.m


config = {
    'tables': {
        'mytable': {
            'lengths': (random_lengths, [10]),
            'lengths_in_cm': (lengths_in_cm, '$mytable.lengths')}}}

pipeline = Pipeline(config)
pipeline.execute()

print(pipeline.state)
```

```shell
$ python test.py
python test.py
Traceback (most recent call last):
  File "test.py", line 21, in <module>
    pipeline.execute()
  File "skypy/pipeline/_pipeline.py", line 193, in execute
    self.state[table][column] = self.get_value(settings)
  File "skypy/pipeline/_pipeline.py", line 245, in get_value
    result = function(parsed_args)
  File "test.py", line 7, in lengths_in_cm
    return q.to_value(units.cm)
AttributeError: 'Column' object has no attribute 'to_value'
```

**Expected behavior**
Table columns with units should be passed in a `Quantity`-like manner.

**Additional context**
This should ideally be fixed inside Astropy, but either way, we need to work around this issue for compatibility with Astropy 4.0 one way or another -- where the "another" way is to only use the subset of `Quantity` methods that `Column` supports, e.g. no `to_value`.

rrjbca:
## Description
`Pipeline.__getitem__` currently retrieves column data as an `astropy.table.Column` object. This is problematic as many of our functions are written to expect `astropy.units.Quantity` objects as arguments but the `Column` class doesn't support all of the same methods of the `Quantity` class e.g. `to_value`. This pull request modifies `Pipeline.__getitem__` to return column data as an `astropy.units.Quantity` object. Resolves #355. 

## Checklist
- [x] Write unit tests
- [ ] ~Write documentation strings~
- [ ] ~Assign someone from your working team to review this pull request~

### DIFF
--- a/skypy/pipeline/_pipeline.py
+++ b/skypy/pipeline/_pipeline.py
@@ -292,4 +292,7 @@ class Pipeline:
     def __getitem__(self, label):
         name, _, key = label.partition('.')
         item = self.state[name]
-        return item[key].quantity if key else item
+        if key:
+            return item[key].data if item[key].unit is None else item[key].quantity
+        else:
+            return item
--- a/skypy/pipeline/_pipeline.py
+++ b/skypy/pipeline/_pipeline.py
@@ -292,4 +292,4 @@ class Pipeline:
     def __getitem__(self, label):
         name, _, key = label.partition('.')
         item = self.state[name]
-        return item[key] if key else item
+        return item[key].quantity if key else item
--- a/skypy/pipeline/tests/test_pipeline.py
+++ b/skypy/pipeline/tests/test_pipeline.py
@@ -3,6 +3,7 @@ from astropy.cosmology.core import Cosmology
 from astropy.io import fits
 from astropy.table import Table
 from astropy.table.column import Column
+from astropy.units import Quantity
 from astropy.utils.data import get_pkg_data_filename
 import networkx
 import numpy as np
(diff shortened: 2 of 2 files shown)


skypy/pipeline/_pipeline.py
```diff
@@ -292,4 +292,4 @@ def get_deps(self, args):
     def __getitem__(self, label):
         name, _, key = label.partition('.')
         item = self.state[name]
-        return item[key] if key else item
+        return item[key].quantity if key else item
```
JonathanDHarris, line 295: @rrjbca 
I think this runs into problems if the column has a unit type of `None`.

So I think you want something like:
```
        if key:
            return item[key].quantity if item[key].unit else item[key]
        return item
```

rrjbca: Thanks, changed this slightly to return the data as numpy array if there are no units (instead of an astropy column)

ntessore (approved):
Think this is now the best way to handle this.

PR #365, opened 2020-11-12
Raise exception if config keys are not strings

### LINKED ISSUE #364: SkyPyLoader parses names as non-string variables
The SkyPyLoader parses variable names using the same logic that it parses their values. As a result, variable names can take non-string types including int, float and astropy Quantity. Consider the config file:

```yml
# config.yml
42_km: 42 km
43 km: 43 km
```

``` python
>>> from skypy.pipeline import load_skypy_yaml
>>> config = load_skypy_yaml("./config.yml")
>>> print(config)

{'42_km': <Quantity 42. km>, <Quantity 43. km>: <Quantity 43. km>}
```

For the first entry, the key "42_km" is correctly interpreted as a string. However for the second entry  the key "43 km" is interpreted as an astropy Quantity by the same logic that applies to its corresponding value. This is problematic because `Pipeline.__getitem__` assumes all variable names in `state` are strings.

@ntessore pointed out that pyyaml will not let you apply different logic to keys and values when parsing files and therefore suggested that instead we enforce all variable names must be strings within the Pipeline class.

JonathanDHarris:
## Description
Resolves #364 

Raises a `ValueError` if one of the keys in the yaml config is not a String.

## Checklist
- [x] Write unit tests
- [ ] ~Assign someone from your working team to review this pull request~


skypy/pipeline/_config.py
```diff
@@ -30,7 +30,16 @@ def load(cls, stream):
...
+            single_data = loader.get_single_data()
+
+            keys = single_data.keys() if single_data else {}
+            for key in keys:
+                if not isinstance(key, str):
+                    error_message = f"Invalid key found in config.  {key} is not a string." \
+                                    f"  Either rename this value or wrap it in quotes."
+                    raise ImportError(error_message)
+
+            return single_data
```
rrjbca, line 42: Thanks @JonathanDHarris looks good and I agree with the decision to check this in the conifg parser rather than the pipeline. Since the config can contain nested dictionaries of variable definitions we will need to check the keys recursively to ensure all parameters, tables, columns etc also have string names. And while we are at it we should also check that the kwargs for function calls also have string names. I've suggested a recursive implementation based off your code (untested) feel free to use or modify it. If anything is unclear just ask.

```suggestion
            def validate_keys(config):
                for k in config.keys():
                    if not isinstance(k, str):
                        raise ValueError(f"Invalid key found in config. {k} is not a string. "
                                         f"Either rename this value or wrap it in quotes.")
            
            from collections.abc import Mapping
            def validate_config(config):
                # Check each key at the current depth is a string
                validate_keys(config)
                for v in config.values():
                    # If any values are dictionaries, recurse
                    if isinstance(v, Mapping):
                        validate_config(v)
                    # If any values are tuples (i.e. function calls) validate kwargs
                    if isinstance(v, tuple) and len(v) > 1 and isinstance(v[1], Mapping):
                        validate_keys(v[1])
                return config

            return validate_config(single_data if single_data else {})
```

JonathanDHarris:
Thanks @rrjbca .  I moved the import and the methods off the `SkyPyLoader` class.  I was trying to neaten it up but I have to be honest- I don't entirely know what the preferred style is for Python and I was partly following PyCharm's suggestions when I did this.  I can move them back how you wrote them if that's correct.

rrjbca (commented):
Thanks @JonathanDHarris, looks good to me, just making a few cosmetic changes then I'll merge

rrjbca (approved):

Issue #318, opened 2020-10-05
Optimising parametrisation in config file


philipp128:
## Description
Enable to have an easier option for the parametrisation in the config file.

## Current Problem
 Let's look at the case that we want to sample galaxies from the Schechter function. An example config file is `mccl_galaxies.yml` in `skypy/examples`.
The model of `phi_star` is given as an exponential function `b*exp(a*z)` where `b` and `a` are the model parameters. 
At the moment we are using Astropy's `astropy.modeling.models.Exponential1D` function to parametrise this model. However, this requires the input parameter `tau` which corresponds to `1/a`. 

As usually the parameter `a` is given in the literature the user has to calculate the parameter `tau` by hand and add it to the config file. A more user-friendly way might be good for the future.

## Suggestion
 - Option to include lambda functions in the config file to construct your own models
- Enable calculations in the config file such that it is possible to write `tau` as `1/a`

PR #374, opened 2020-11-20
Lambda functions can be defined in config

JonathanDHarris:
## Description
Allow users to define lambda functions in their configs.  Resolve's #318 

## Checklist
- [ ] Write unit tests

### DIFF
--- a/skypy/pipeline/_config.py
+++ b/skypy/pipeline/_config.py
@@ -81,6 +81,14 @@ class SkyPyLoader(yaml.SafeLoader):
 
         return (function, args, kwargs)
 
+    def construct_lambda(self, node):
+
+        lambda_val = node.value
+
+        assign_lambda_to_temp = 'temp = lambda {}'.format(lambda_val)
+        exec(assign_lambda_to_temp, globals())
+        return temp
+
     def construct_quantity(self, node):
         value = self.construct_scalar(node)
         return Quantity(value)
@@ -88,6 +96,9 @@ class SkyPyLoader(yaml.SafeLoader):
 
 # constructor for generic functions
 SkyPyLoader.add_multi_constructor('!', SkyPyLoader.construct_function)
+
+# constructor for lambda functions
+SkyPyLoader.add_constructor('!lambda', SkyPyLoader.construct_lambda)
 
 # constructor for quantities
 SkyPyLoader.add_constructor('!quantity', SkyPyLoader.construct_quantity)
--- a/skypy/pipeline/tests/data/lambda_function.yml
+++ b/skypy/pipeline/tests/data/lambda_function.yml
@@ -1,0 +1,3 @@
+a: 0.5
+tau: !lambda 'a : 1. / a'
+twice: !lambda 'a : a * 2'
--- a/skypy/pipeline/tests/test_config.py
+++ b/skypy/pipeline/tests/test_config.py
@@ -65,3 +65,11 @@ def test_kwarg_must_be_strings():
     with pytest.raises(ValueError) as e:
         load_skypy_yaml(filename)
     assert("Invalid key found in config" in e.value.args[0])
+
+
+def test_lambda_functions():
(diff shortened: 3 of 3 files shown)


ntessore:
I have a very similar changeset on my machine, which I never propose because of the dangers of `eval()`. My idea was to only eval the lambda function body:

```py
def = 'x**2 + 3'
f = lambda x: eval(def)
```

I think this would severely limit the harm that can be done in the lambda function body. But then I never figured out a good way to specify the function arguments.

JonathanDHarris:
I'm not sure there's anyway to resolve #318 without using `eval`, so I think a decision needs to be made to use `eval` or close the issue off as won't fix.

ntessore:
It's true (at the moment) that arbitrary functions can be run from config files, and that's another issue. But I wasn't thinking of malicious code `rm -rf`ing the system as much as someone doing something weird that messes up SkyPy internals. Hence why I would prefer to limit the input to be only a single statement.

There is a way to build lambdas programmatically, but it requires work:

```py
>>> lambda_args = ['x', 'y']
>>> lambda_body = 'x**2 + y + 1'
>>>
>>> import ast
>>> args = ast.arguments(posonlyargs=[],
...                      args=[ast.arg(arg) for arg in lambda_args],
...                      kwonlyargs=[],
...                      kw_defaults=[],
...                      defaults=[])
>>> body = ast.parse(lambda_body, mode='eval').body
>>> expr = ast.Expression(ast.Lambda(args, body))
>>> expr.line_no = 1  # from YAML
>>> expr.col_offset = 1  # from YAML
>>> expr = ast.fix_missing_locations(expr)
>>> code = compile(expr, filename='<ast>', mode='eval')
>>> func = eval(code, {})
>>> func(2, 3)
8
```

And this still leaves the problem of how to obtain the list of arguments and the function body from the YAML.

JonathanDHarris:
"And this still leaves the problem of how to obtain the list of arguments and the function body from the YAML."

Something like this should work
```
function_def: !lambda
  x: 1
  b: 2
  function_def:'x**2 + y + 1'
```

ntessore:
We want the lambdas to produce callables and not values, in order to define parametrisations such as [in the examples](https://github.com/skypyproject/skypy/blob/master/examples/mccl_galaxies.yml#L7-L8) in a flexible manner.

Since there is no good way to use YAML intrinsics to make this easier, I think the `!lambda` tag should simply take its value as a scalar node (string) and parse. If we are using Python lambda syntax:

```
mycallable: !lambda 'x, y: x**2 + y + 1'
```

Since there's a `:` in the syntax, we have no way around quotes. However, I have always been fond of Julia's anonymous function syntax:

```
mycallable: !lambda (x, y) -> x**2 + y + 1
```

In addition to all of this, we will also need to export symbols such as `exp`, `sin`, `cos`, etc. for the evaluation of the function body. That's quite the big task, so don't feel obligated to tackle this issue.

ntessore:
On further thought there's even a whole other infrastructure dimension to this, because the lambda's should have access to certain other parts of the pipeline, such as parameters. @JonathanDHarris I will review this now and then we will iterate later on.

skypy/pipeline/_config.py
```diff
@@ -81,6 +81,14 @@ def construct_function(self, name, node):
 
         return (function, args, kwargs)
 
+    def construct_lambda(self, node):
+
+        lambda_val = node.value
+
+        assign_lambda_to_temp = 'temp = lambda {}'.format(lambda_val)
+        exec(assign_lambda_to_temp, globals())
+        return temp
```
ntessore, line 90:
```suggestion
        lambda_val = 'lambda ' + self.construct_scalar(node)
        return eval(lambda_val, {})
```

Simpler construction that also warns if a sequence or mapping node is given.

ntessore (changes requested):

ntessore (approved):

PR #378, opened 2020-11-21
pipeline: generic interface for pipeline items

ntessore:
Provide a generic base class `Item` for pipeline items, and two concretisations `Ref` (reference to another variable) and `Call` (function call). The pipeline items have three methods:

- `infer` to infer additional properties from context (e.g. the `cosmology` argument in `Call`),
- `depend` to return the dependencies of the item, and
- `evaluate` to return the computed value of the item.

This removes special cases from the `Pipeline` class and makes it easily extendable to other types of items.


skypy/pipeline/tests/test_config.py
```diff
@@ -34,6 +32,12 @@ def test_load_skypy_yaml():
         load_skypy_yaml(filename)
 
 
+def test_empty_ref():
+    filename = get_pkg_data_filename('data/test_empty_ref.yml')
+    with pytest.raises(ValueError, match='empty reference'):
```
JonathanDHarris, line 37: Looking at the yml file for this test, is this still an empty reference?  Or a broken one?

ntessore: My thinking here was that at the config (i.e. YAML) level we can only catch empty strings being passed when the `!ref` tag is used without argument. We could also let it go to the pipeline level and check there, but then we lose the information about where the error happened.

Which now makes me think that the next iteration of this should be adding 'file', 'line' and 'column' information to `Item` on construction so that the pipeline can do all of the checking and give great error messages.

skypy/pipeline/_pipeline.py
```diff
@@ -256,53 +209,42 @@ def write(self, file_format=None, overwrite=False):
...
+        if isinstance(args, Sequence) and not isinstance(args, str):
+            # recurse list
+            return sum([self.depend(a) for a in args], [])
+        elif isinstance(args, Mapping):
             # get explicit dependencies
             deps = args.pop('.depends', [])
             # turn a single value into a list
-            if isinstance(deps, str) or not isinstance(deps, list):
+            if isinstance(deps, str) or not isinstance(deps, Sequence):
                 deps = [deps]
```
rrjbca, line 242:
```suggestion
            if isinstance(deps, str) or not isinstance(deps, Sequence):
                deps = Ref(deps).depend(self)
```
I think this does the same thing but wonder if it makes it clearer?

skypy/pipeline/tests/test_items.py
```diff
@@ -0,0 +1,63 @@
...
+    # construct the base class
+    item = Item()
+
+    # check `infer` method
+    item.infer({})
+
+    # check `depend` method
+    deps = item.depend(pipeline)
+    assert isinstance(deps, list)
+
```
rrjbca, line 20:
```suggestion
    assert len(deps) == 0

```
A stronger check, since we know base Item has no dependencies i.e. an empty list?

skypy/pipeline/tests/test_items.py
```diff
@@ -0,0 +1,63 @@
...
+    # check `evaluate` method
+    val = item.evaluate(pipeline)
+    assert val is None
+
+
+def test_call():
```
rrjbca, line 26: Could we modify this test slightly to provide one of the arguments as a pipeline variable (instead of in context) and then check that Call.depends returns the appropriate Ref?

rrjbca (changes requested):

skypy/pipeline/_pipeline.py
ntessore: I like that at the `Pipeline` level, all refs are `Ref` objects and not strings. But since this changes YAML syntax, let's do it in a separate changes.

rrjbca (approved):

PR #376, opened 2020-11-20
Installation instructions in documentation

rrjbca:
## Description
As suggested in #354 move the installation instructions from the README into a new *Installation* page of the documentation to reduce clutter. 

- [Front Page](https://skypy--376.org.readthedocs.build/en/376/)
- [Installation Page](https://skypy--376.org.readthedocs.build/en/376/install.html)

## Checklist
- [x] Follow the [Contributor Guidelines](https://github.com/skypyproject/skypy/blob/master/CONTRIBUTING.md)
- [ ] ~Write unit tests~
- [ ] ~Write documentation strings~
- [ ] ~Assign someone from your working team to review this pull request~
- [x] Assign someone from the infrastructure team to review this pull request

PR #379, opened 2020-11-23
Add kcorrect template data to SkyPy package

rrjbca:
## Description
Following from discussion in #376, this pull request moves the kcorrect template data from the [skypyproject/skypy-data](https://github.com/skypyproject/skypy-data) repository into the main SkyPy package and implements a loader function. Together with #368 we will be able to drop skypy-data as a dependency and deprecate that repository.

## Checklist
- [ ] Write unit tests


ntessore:
Ah, good, I had started on this as well. In case interesting, my thoughts were:

- To Gzip the data files to keep the impact on file size low, and
- To remove the generic lookup and loading and make the loader function only look in the `spectrum_templates` folder for the named file.

rrjbca:
- Does this really make much difference? I'm not sure I can be bothered haha
- I was going to merge #368 first and then refactor some of the code, perhaps a SpectrumLoader base class with SpecliteLoader and SkyPyDataLoader subclasses? Or yes your suggestion would also be similarly sensible

ntessore:
It's a 60% reduction in size with one line of bash. The loading of gzipped files happens transparently.

```shell
skypy-data$ du -sh spectrum_templates/
5.7M	spectrum_templates/
skypy-data$ for f in spectrum_templates/*.ecsv; do gzip $f; done
skypy-data$ du -sh spectrum_templates/
2.1M	spectrum_templates/
```

setup.cfg
```diff
@@ -45,7 +45,7 @@ docs =
     matplotlib
 
 [options.package_data]
-skypy = data/*,*/tests/data/*
+skypy = data/*,data/spectrum_templates/*,*/tests/data/*
```
ntessore, line 48: Would prefer this for if we ever add other data.

```suggestion
skypy = data/*,data/*/*,*/tests/data/*
```

skypy/galaxy/_spectrum_loaders.py
```diff
@@ -94,6 +94,39 @@ def skypy_data_loader(module, name, *tags):
...
+
+    # result is spectrum or list of spectra
+    spectra = None
+
+    # load each tag separately
+    for tag in tags:
+
+        # load the data from skypy/data
+        path = astropy.utils.data._find_pkg_data_path(f'data/{module}', 'skypy')
+        filename = os.path.join(path, f'{name}_{tag}.ecsv')
```
ntessore, line 108: Would prefer not using a private Astropy function (or any Astropy function for that matter).
```suggestion
        filename = resource_filename('skypy', f'data/{module}/{name}_{tag}.ecsv')
```

ntessore (commented):
I still think it better to gzip these files so that we don't have a +80k line changeset, but that's up to you.

ntessore (dismissed):

ntessore (approved):

PR #392, opened 2021-01-20
remove module dependencies

ntessore:
Remove dependencies from master that belong to stripped modules (#389).


philipp128 (approved):
As I can see it only contains the dependencies necessary for the galaxy module as supposed to do.

rrjbca (approved):

PR #415, opened 2021-01-26
magnitude/luminosity functions to utils.photometry

ntessore:
## Description

Move the magnitude and luminosity conversion functions to the new `skypy.utils.photometry` module. Removes the incorrect apparent and absolute magnitude conversion functions.


docs/pipeline/examples/config.yml
```diff
@@ -4,12 +4,11 @@ parameters:
...
-n_galaxies: 100
+n_galaxies: 1000
 tables:
   galaxies:
     redshift: !skypy.galaxies.redshift.smail
       z_median: 1.0
       alpha: 1.5
       beta: 2.0
       size: $n_galaxies
-    distance_modulus: !skypy.galaxies.luminosity.distance_modulus [$galaxies.redshift]
```
rrjbca, line 15: How about: `distance_modulus: !astropy.cosmology.Planck15.distmod [$galaxies.redshift]`? I note that we don't have a way to call `distmod` for the user defined cosmology in this pipeline and that is something we should develop in the future, but at least this way we can still demonstrate how to reference a column (`$galaxies.redshift`)

ntessore: I agree with the sentiment but that makes a horribly inconsistent example. There must be a better consumer of `$galaxies.redshift`.

rrjbca: I looked but couldn't find one. OK I will merge this as is

rrjbca (approved):

PR #422, opened 2021-02-05
ADR 3: Position sampling and patches of the sky

ntessore:
ADR for the sampling of positions on the sky. Direct link to proposal [here](https://github.com/ntessore/skypy/blob/adr-03/docs/arch/adr-03.md).

### DIFF
--- a/docs/arch/adr-03.md
+++ b/docs/arch/adr-03.md
@@ -113,3 +113,8 @@ list of regions. For example, a very real use case would be to generate
 populations of galaxies and simply count the total number in each HEALPix pixel
 to generate a density map. This will be addressed in a separate ADR.
 
+
+Consequences
+------------
+The existing `Pipeline` class must be extended to support iterating regions. No
+existing interfaces are affected.
--- a/docs/arch/adr-03.md
+++ b/docs/arch/adr-03.md
@@ -4,7 +4,7 @@ ADR 3: Position sampling and patches of the sky
 
 Author: Nicolas Tessore  
 Date: 5 February 2021  
-Status: draft
+Status: Accepted
 
 
 Context
--- a/docs/arch/adr-03.md
+++ b/docs/arch/adr-03.md
@@ -1,0 +1,115 @@
+
+ADR 3: Position sampling and patches of the sky
+===============================================
+
+Author: Nicolas Tessore  
+Date: 5 February 2021  
+Status: draft
+
+
+Context
+-------
+
+This ADR addresses two related open problems for SkyPy simulations:
+
+- How can we describe the regions in which to sample the positions of e.g.
+  galaxies or supernovae?
+- How can we break up the sampling over large fractions of the sky into smaller
+  chunks?
+
+The second point is particularly relevant when the simulation becomes so large
+that it no longer fits into memory, as well as for parallelisation.
+
+
+Extension to pipelines
+----------------------
+
+This ADR proposes to introduce a new top-level keyword `regions` [alt: `sky`,
+`geometry`, `patches`] into pipelines that describes the geometry of the
+simulation. For example, to include a "rectangular" and a "circular" region:
+
+```yaml
+regions:
+- !rectangular [ ... ]
+- !circular [ ... ]
+```
+
+The `regions` list can be traversed by the pipeline runner to create what are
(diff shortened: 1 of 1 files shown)


docs/arch/adr-03.md
```diff
@@ -0,0 +1,115 @@
...
+----------------------
+
+This ADR proposes to introduce a new top-level keyword `regions` [alt: `sky`,
+`geometry`, `patches`] into pipelines that describes the geometry of the
+simulation. For example, to include a "rectangular" and a "circular" region:
+
+```yaml
+regions:
+- !rectangular [ ... ]
+- !circular [ ... ]
```
CKrawczyk, line 34: Can you clarify what would go inside the `[ ... ]`, I am unsure what this would look like.  Would it be a bounding box for the rectangle, what coordinate system would it use, etc...

docs/arch/adr-03.md
```diff
@@ -0,0 +1,115 @@
...
+regions: !healpix
+  mask: survey-footprint-n512.fits
+```
+
+The `mask` keyword can be combined with the `nside` parameter to change the
+resolution of the mask if requested.
+
+If the HEALPix maps become finely resolved, it may be desirable to combine
+several pixels into a single region. There may be a `batch` [alt: `combine`]
+keyword for this purpose:
```
CKrawczyk, line 95: How will this `batch` work exactly?  Is it just resampling to a lower res healpix grid or some other way?  Does that mean `batch` needs to be a power of 2?  If so might be better off with `nside_batch` (there has to be a better keyword that could be used) or similar.

docs/arch/adr-03.md
CKrawczyk: Although that might be a topic for a different ADR that outlines how the regions are made.

docs/arch/adr-03.md
```diff
@@ -0,0 +1,115 @@
...
+the current region in a `$region` reference that can be used where necessary.
+For example, to sample from a luminosity function with positions:
+
+```yaml
+tables:
+  galaxies:
+    z, M: !schechter_lf
+      ...
+      sky_area: $region.area
+    ra, dec: !random_point [ $region ]
```
CKrawczyk, line 63: I am a bit out of date with how SkyPy config files work, is this the same `yaml` file as above or a separate file?

CKrawczyk (changes requested):
I have pointed out some points that can be clarified above.  It would also be nice if this ADR added: 
- "Status = pending" section that is updated to "accepted" when the review is done, and "implemented" when it makes it into the code
- a section for "Consequences" that goes into how these changes would affect the existing code (could be "does not affect existing code").

docs/arch/adr-03.md
ntessore: Yes, these are only placeholders, there would likely be a couple of functions taking different specifications.

docs/arch/adr-03.md
ntessore: That would be up to the implementation. Since there is no good way of sampling HEALPix pixels besides rejection sampling from a larger region, any number of schemes are possible. It will be easier to decide this when it becomes necessary.

docs/arch/adr-03.md
ntessore: This is the same file. The top-level region specification is looped over in the pipeline class, and each loop creates essentially a table fragment as specified here.

docs/arch/adr-03.md
```diff
@@ -0,0 +1,115 @@
...
+
+
+Map making
+----------
+
+This ADR does not address the problem of how maps will be generated from the
+list of regions. For example, a very real use case would be to generate
+populations of galaxies and simply count the total number in each HEALPix pixel
+to generate a density map. This will be addressed in a separate ADR.
+
```
ntessore, line 115:
```suggestion


Consequences
------------
The existing `Pipeline` class must be extended to support iterating regions. No
existing interfaces are affected.

```

CKrawczyk (dismissed):
Looks good to me.

rrjbca (dismissed):
I like this design. The implementation of e.g. region batches will require some thought but the proposal seems flexible enough to accommodate this.

rrjbca (approved):

PR #441, opened 2021-03-01
Update status badges

rrjbca:
## Description
See updated README [here](https://github.com/skypyproject/skypy/blob/d63907ac0611e9212c8db3e7ad6405deedd07838/README.rst)
- Add JOSS status badge
- Move status badges to appropriate subsections in the README to reduce clutter

## Checklist
- [ ] Write unit tests

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -2,7 +2,7 @@
 SkyPy: A package for modelling the Universe
 ===========================================
 
-|PyPI| |conda-forge| |Read the Docs| |GitHub| |Codecov| |JOSS| |Zenodo|
+|Read the Docs| |GitHub| |Codecov|
 
 This package contains methods for modelling the Universe, galaxies and the Milky
 Way. Also included are methods for generating observed data.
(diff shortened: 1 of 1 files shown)


ntessore:
- no point in adding the JOSS badge unless it's accepted
- I get the urge to remove clutter, but the entire point of the badges is to have them at the top of the README for a quick glance. Instead of moving them you might as well remove them and replace with a blurb of text.

rrjbca:
The JOSS badge would update on being accepted (which I recognise could not happen). Regarding clutter the specific issue I had was that adding JOSS made the badges spill over onto a second line on my screen. They do also seem to split naturally into groups; DOIs, Distributions, and Testing (+Docs?). Probably preferable to keep the DOI badges on the front page though rather than getting lost in the CITATION page.

rrjbca:
Perhaps let's keep the PR open and reconsider merging once the JOSS paper is published?

ntessore:
In the end, now with the badges under individual headlines, it is easily identifiable [on the actual page](https://github.com/skypyproject/skypy/blob/d63907ac0611e9212c8db3e7ad6405deedd07838/README.rst) so happy to accept once JOSS goes through. Or we accept it now and own our loss in case it doesn't get accept.

ntessore (approved):
I decided that I like the Under Review badge, so good to go for me.

PR #455, opened 2021-03-29
ENH: Power spectrum base classes

rrjbca:
## Description
Implements PowerSpectrum (and TabulatedPowerSpectrum) classes to provide a common interface for power spectrum calculations (using interpolation). Also implements EisensteinHu, CAMB and CLASS classes to replace previous raw functions.

N.b. data used when testing CLASS was found to be inaccurate and has been recalculated. However, even the new "correct" data does not match our interpolation method at the previous relative tolerance of 1e-4 so this has been reduced to 1e-3.

## Checklist
- [x] Write unit tests

### DIFF
--- a/skypy/power_spectrum/_base.py
+++ b/skypy/power_spectrum/_base.py
@@ -1,0 +1,23 @@
+from abc import ABCMeta, abstractmethod
+import numpy as np
+from scipy.interpolate import RectBivariateSpline
+
+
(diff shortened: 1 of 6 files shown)


jucordero (approved):
Looks good. Probably better to just go ahead so we can start with the next steps right away

itrharrison (approved):
LGTM

PR #489, opened 2021-09-03
TST: Use tmp_path fixture for temporary files in unit tests

### LINKED ISSUE #466: FITS file deletion fails when tests fail
When unit tests that create output FITS files fail, the test cleanup code might try to delete nonexisting files, causing other errors that are unrelated to what's actually failing in the test.

Instead of using the cleanup function, we can use a pytest `tmp_path` fixture to write files to a pytest temporary directory.

From their [tutorial](https://docs.pytest.org/en/latest/how-to/tmpdir.html):

```py
def mytest(tmp_path):
    out_path = tmp_path / 'test.fits'
```

rrjbca:
## Description
Use pytest `tmp_path` fixture in unit tests. Creates a temporary directory where temporary files can be written and handles cleanup automatically. Resolves #466 

## Checklist
- [ ] Write unit tests

### DIFF
--- a/skypy/pipeline/tests/test_pipeline.py
+++ b/skypy/pipeline/tests/test_pipeline.py
@@ -21,7 +21,7 @@ else:
     HAS_H5PY = True
 
 
-def test_pipeline():
+def test_pipeline(tmp_path):
 
     # Evaluate and store the default astropy cosmology.
     config = {'test_cosmology': Call(default_cosmology.get)}
@@ -44,7 +44,7 @@ def test_pipeline():
 
     pipeline = Pipeline(config)
     pipeline.execute()
-    output_filename = 'output.fits'
+    output_filename = str(tmp_path / 'output.fits')
     pipeline.write(output_filename)
     assert len(pipeline['test_table']) == size
     assert np.all(pipeline['test_table.column1'] < pipeline['test_table.column2'])
@@ -251,7 +251,7 @@ def test_column_quantity():
 
 
 @pytest.mark.skipif(not HAS_H5PY, reason='Requires h5py')
-def test_hdf5():
+def test_hdf5(tmp_path):
     size = 100
     string = size*'a'
     config = {'tables': {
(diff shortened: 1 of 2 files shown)


skypy/pipeline/tests/test_skypy.py
```diff
@@ -36,36 +35,39 @@ def test_skypy():
     assert e.value.code == 2
 
     # Process empty config file
-    filename = get_pkg_data_filename('data/empty_config.yml')
-    assert skypy.main([filename, 'empty.fits']) == 0
+    config_filename = get_pkg_data_filename('data/empty_config.yml')
```
Lucia-Fonseca, line 38: why don't we use a temporary YAML file here too?

Lucia-Fonseca (commented):
The code fixes the problem.
However, I wonder whether `unittest.mock` is more suitable for mocking files, so we don't need to create temporary files or have empty yml files(?)

skypy/pipeline/tests/test_pipeline.py
```diff
@@ -44,7 +43,7 @@ def test_pipeline():
 
     pipeline = Pipeline(config)
     pipeline.execute()
-    output_filename = 'output.fits'
+    output_filename = str(tmp_path / 'output.fits')
```
philipp128, line 46: Do we need the `str` command? It is not used in the [tutorial](https://docs.pytest.org/en/6.2.x/tmpdir.html)

rrjbca: "`tmp_path` is a `pathlib.Path` object" but our script takes path names as strings

rrjbca:
> However, I wonder whether `unittest.mock` is more suitable for mocking files, so we don't need to create temporary files or have empty yml files(?)

This PR is not mocking any files. It is implementing a more robust method (a temporary directory) to automatically handle the deletion of files created by the skypy script during testing. We don't want to mock those files; the fundamental purpose of these unit tests is to check that those files are created correctly. Mocking config files might be possible but is unrelated to the issue this PR is addressing and I don't see an obvious benefit?

skypy/pipeline/tests/test_skypy.py
rrjbca: This is a (non-temporary) input file that is part of the repository

Lucia-Fonseca (approved):

Issue #513, opened 2022-02-17
ENH: Quenching model


Lucia-Fonseca:
## Description
Add a feature where we can obtain the Schechter parameters for active galaxies (centrals and satellites) and passive galaxies (mass- and satellite-quenched) based on equation (15) in de la Bella et al. 2021.
## Inputs
 - Active Schechter parameters
 - Fraction of satellite galaxies
 - Fraction of satellite-quenched galaxies

## Outputs
 - Schechter parameters for active galaxies (centrals and satellites)
 - Schechter parameters for passive galaxies (mass- and satellite-quenched)

## References
* de la Bella et al. 2021, Quenching and Galaxy Demographics, [arXiv 2112.11110](https://arxiv.org/pdf/2112.11110.pdf).

PR #516, opened 2022-02-17
ENH: galaxy-quenching model

Lucia-Fonseca:
## Description
This PR adds a feature where we can obtain the Schechter parameters for active galaxies (centrals and satellites) and passive galaxies (mass- and satellite-quenched) based on equation (15) in de la Bella et al. 2021. Merging this PR coses #513 .

## Checklist
- [ ] Follow the [Contributor Guidelines](https://github.com/skypyproject/skypy/blob/main/CONTRIBUTING.rst)
- [ ] Write unit tests

### DIFF
--- a/skypy/galaxies/stellar_mass.py
+++ b/skypy/galaxies/stellar_mass.py
@@ -8,6 +8,7 @@ from ..utils.random import schechter
 
 __all__ = [
     'schechter_smf_mass',
+    'schechter_smf_parameters',
 ]
 
 
@@ -74,3 +75,54 @@ def schechter_smf_mass(redshift, alpha, m_star, m_min, m_max, size=None,
     m = schechter(alpha, x_min, x_max, resolution, size=size, scale=m_star)
 
     return m
+
+
+def schechter_smf_parameters(active_parameters, fsatellite, fenvironment):
+    r'''Schechter parameters.
+    This function returns the Schechter mass function parameters
+    for active galaxies (centrals and satellites)
+    and passive galaxies (mass- and satellite-quenched)
+    based on equation (15) in de la Bella et al. 2021 [1]_.
+
+    Parameters
+    ----------
+    active_parameters: tuple
+        Schechter mass function parameters for the entire active
+        sample of galaxies: :math:`(\phi_b, \alpha_b, m_{*})`.
+
+    fsatellite: float, (n, ) array_like
+        Fraction of active satellite galaxies between 0 and 1.
+        It could be a float or an array, depending on the model you choose.
+
+    fenvironment: float
+        Fraction of satellite-quenched galaxies between 0 and 1.
+
+
+    Returns
+    -------
+    parameters: dic
+        It returns a  dictionary with the parameters of the
+        Schechter mass function. The dictionary keywords:
+        `centrals`, `satellites`, `mass_quenched` and
+        `satellite_quenched`. The values correspond to a
+        tuple :math:`(\phi, \alpha, m_{*})`.
+
+    References
+    ----------
+    .. [1] de la Bella et al. 2021, Quenching and Galaxy Demographics,
+           arXiv 2112.11110.
+
+    '''
+    phi, alpha, mstar = active_parameters
+
+    sum_phics = (1 - fsatellite) * (1 - np.log(1 - fsatellite))
+    phic = (1 - fsatellite) * phi / sum_phics
+    phis = phic * np.log(1 / (1 - fsatellite))
+
+    centrals = (phic, alpha, mstar)
+    satellites = (phis, alpha, mstar)
+    mass_quenched = (phi, alpha + 1, mstar)
+    satellite_quenched = (- np.log(1 - fenvironment) * phis, alpha, mstar)
+
+    return {'centrals': centrals, 'satellites': satellites,
+            'mass_quenched': mass_quenched, 'satellite_quenched': satellite_quenched}
--- a/docs/galaxies.rst
+++ b/docs/galaxies.rst
@@ -89,6 +89,7 @@ The following models are found in the `skypy.galaxies.stellar_mass` package.
   :nosignatures:
 
   schechter_smf_mass
+  schechter_smf_parameters
 
 
 Reference/API
--- a/skypy/halo/quenching.py
+++ b/skypy/halo/quenching.py
@@ -1,0 +1,106 @@
+"""Galaxy quenching.
+
+This module implements models for environment and mass
+quenching by dark matter halos.
+"""
+
+import numpy as np
+from scipy import special
+
+__all__ = [
+    'environment_quenched',
+    'mass_quenched',
+    ]
+
+
+def environment_quenched(nh, probability):
+    r'''Environment quenching function.
+    This function implements the model proposed by A.Amara where the
+    probability of a subhalo being quenched is a fixed
+    probability. The model is inspired on [1]_ and [2]_.
+
+    Parameters
+    ----------
+    nh: integer
+        Number of subhalos.
+    probability: float
+        Quenching probability.
+
+    Returns
+    -------
+    quenched: (nh,) array_like,  boolean
+        Boolean array indicating which subhalo's host galaxies are
+        (satellite) environment-quenched.
+
+    Examples
+    ---------
+
+    This example shows how many subhalos are environtment quenched (True)
+    and how many survive (False) from a list of 1000 halos:
+
+    >>> import numpy as np
+    >>> from skypy.halo.quenching import environment_quenched
+    >>> from collections import Counter
+    >>> quenched = environment_quenched(1000, 0.5)
+    >>> Counter(quenched)
+    Counter({...})
+
+    References
+    ----------
+    .. [1] Peng et al. 2010, doi 10.1088/0004-637X/721/1/193.
+    .. [2] Birrer et al. 2014, arXiv 1401.3162.
+
+    '''
+
+    return np.random.uniform(size=nh) < probability
+
+
+def mass_quenched(halo_mass, offset, width):
+    r'''Mass quenching function.
+    This function implements the model proposed by A.Amara where the
+    probability of a halo being quenched is related to the error function
+    of the logarithm of the halo's mass standardised by an offset and width
+    parameter.  The model is inspired on [1]_ and [2]_.
+
+    Parameters
+    ----------
+    halo_mass: (nh,) array_like
+        Array of halo masses in units of solar mass, :math:`M_{sun}`.
+    offset: float
+        Halo mass in :math:`M_{sun}` at which quenching probability is 50%.
+    width: float
+        Width of the error function.
+
+    Returns
+    -------
+    quenched: (nh,) array_like, boolean
+        Boolean array indicating which halo's host galaxies are mass-quenched.
+
+    Examples
+    ---------
+
+    This example shows how many halos are mass quenched (True)
+    and how many survive (False) from a list of 1000 halos:
+
+    >>> import numpy as np
+    >>> from astropy import units
+    >>> from skypy.halo.quenching import mass_quenched
+    >>> from collections import Counter
+    >>> offset, width = 1.0e12, 0.5
+    >>> halo_mass = np.random.lognormal(mean=np.log(offset), sigma=width,
+    ...                                 size=1000)
+    >>> quenched = mass_quenched(halo_mass, offset, width)
+    >>> Counter(quenched)
+    Counter({...})
+
+    References
+    ----------
+    .. [1] Peng et al. 2010, doi 10.1088/0004-637X/721/1/193.
+    .. [2] Birrer et al. 2014, arXiv 1401.3162.
+
+    '''
+
+    standardised_mass = np.log10(halo_mass / offset) / width
+    probability = 0.5 * (1.0 + special.erf(standardised_mass/np.sqrt(2)))
+    nh = len(halo_mass)
+    return np.random.uniform(size=nh) < probability
--- a/skypy/halo/tests/test_quenching.py
+++ b/skypy/halo/tests/test_quenching.py
@@ -1,0 +1,30 @@
+import numpy as np
+from scipy import stats
+from collections import Counter
+
+from skypy.halo.quenching import environment_quenched, mass_quenched
+
+
+def test_environment_quenched():
+    # Test the quenching process follows a binomial distribution
+    n, p = 1000, 0.7
+    quenched = environment_quenched(n, p)
+    number_quenched = Counter(quenched)[True]
+
+    p_value = stats.binom_test(number_quenched, n=n, p=p,
+                               alternative='two-sided')
+    assert p_value > 0.05
+
+
+def test_mass_quenched():
+    # Test the quenching process follows a binomial distribution if the
+    # logarithmic halo masses are symmetrical about the offset
+    n = 1000
+    offset, width = 1.0e12, 0.5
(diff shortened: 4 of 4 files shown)


skypy/galaxies/stellar_mass.py
```diff
@@ -74,3 +75,58 @@ def schechter_smf_mass(redshift, alpha, m_star, m_min, m_max, size=None,
     m = schechter(alpha, x_min, x_max, resolution, size=size, scale=m_star)
 
     return m
+
+
+def schechter_smf_parameters(active_parameters, fsatellite, fenvironment):
```
philipp128, line 80: I know a bit picky but maybe change the order of `fsatellite` and `fenvironment` to match the order in equation (15) of your paper. You do not need to change this but just an idea to be consistent.

skypy/galaxies/stellar_mass.py
```diff
@@ -74,3 +75,58 @@ def schechter_smf_mass(redshift, alpha, m_star, m_min, m_max, size=None,
...
+
+
+    Returns
+    -------
+    parameters: dic
+        It returns a  dictionary with the parameters of the
+        Schechter mass function. The dictionary keywords:
+        `centrals`, `satellites`, `mass_quenched` and
+        `satellite_quenched`. The values correspond to a
+        tuple :math:`(\phi, \alpha, m_{*})`.
```
philipp128, line 112: I am confused. Is it meant to return `phi` or `phi_star`? Looking at equation (15) I assumed we want `phi_star`.

skypy/galaxies/stellar_mass.py
```diff
@@ -74,3 +75,58 @@ def schechter_smf_mass(redshift, alpha, m_star, m_min, m_max, size=None,
...
+    This function returns the Schechter mass function parameters
+    for active galaxies (centrals and satellites)
+    and passive galaxies (mass- and satellite-quenched)
+    based on equation (15) in de la Bella et al. 2021 [1]_.
+
+    Parameters
+    ----------
+    active_parameters: tuple
+        Schechter mass function parameters for the entire active
+        sample of galaxies: :math:`(\phi_b, \alpha_b, m_{*})`.
```
philipp128, line 91: I suggest you call the parameter phi_{*,b}. I know it does not make a real difference but then it is consistent equation (15) in the paper.
```suggestion
        sample of galaxies: :math:`(\phi_{*,b}, \alpha_b, m_{*})`.
```

skypy/galaxies/stellar_mass.py
```diff
@@ -74,3 +75,58 @@ def schechter_smf_mass(redshift, alpha, m_star, m_min, m_max, size=None,
...
+        Note: in [1], :math:`\alpha_b` and :math:`m_{*}` are
+        constants. The elements of the tuple should be scalars.
+        Exception: :math:`\phi_b` can be array only when
```
philipp128, line 94: Here the same as above.

skypy/galaxies/stellar_mass.py
```diff
@@ -74,3 +75,58 @@ def schechter_smf_mass(redshift, alpha, m_star, m_min, m_max, size=None,
...
+        `satellite_quenched`. The values correspond to a
+        tuple :math:`(\phi, \alpha, m_{*})`.
+
+    References
+    ----------
+    .. [1] de la Bella et al. 2021, Quenching and Galaxy Demographics,
+           arXiv 2112.11110.
+
+    '''
+    phi, alpha, mstar = active_parameters
```
philipp128, line 120:
```suggestion
    phi_star, alpha, mstar = active_parameters
```

skypy/galaxies/stellar_mass.py
```diff
@@ -74,3 +75,58 @@ def schechter_smf_mass(redshift, alpha, m_star, m_min, m_max, size=None,
...
+
+    sum_phics = (1 - fsatellite) * (1 - np.log(1 - fsatellite))
+    phic = (1 - fsatellite) * phi / sum_phics
+    phis = phic * np.log(1 / (1 - fsatellite))
```
philipp128, line 124:
```suggestion
    phic_star = (1 - fsatellite) * phi / sum_phics
    phis_star = phic * np.log(1 / (1 - fsatellite))
```

skypy/galaxies/stellar_mass.py
```diff
@@ -74,3 +75,58 @@ def schechter_smf_mass(redshift, alpha, m_star, m_min, m_max, size=None,
...
+
+    centrals = (phic, alpha, mstar)
+    satellites = (phis, alpha, mstar)
+    mass_quenched = (phic + phis, alpha + 1, mstar)
+    satellite_quenched = (- np.log(1 - fenvironment) * phis, alpha, mstar)
```
philipp128, line 129: Changes according to above comments

skypy/galaxies/tests/test_stellar_mass.py
```diff
@@ -60,3 +60,61 @@ def calc_cdf(m):
                                              size=1000, resolution=100)
     p_value = scipy.stats.kstest(sample, calc_cdf)[1]
     assert p_value >= 0.01
+
+
+def test_schechter_smf_parameters():
+    from skypy.galaxies.stellar_mass import schechter_smf_parameters
+    # Check scalar inputs
+    blue = (10**-2.423, 10**10.60, -1.21)
+    fsat_scalar, frho = 0.4, 0.2
```
philipp128, line 69: Maybe be consistent with naming convention in method.

```suggestion
    fsatellite_scalar, fenvironment = 0.4, 0.2
```

philipp128 (changes requested):
For the moment I have mainly comments on naming convention. I found it first hard to figure out what is calculated especially cause of the names in the paper. At the end it does not make a big difference but it was confusing to me.

rrjbca (commented):
I would suggest refactoring as follows:
- mstar is not used to calculate any of the phi or alpha values and is returned unchanged in all four populations. I would remove mstar from the arguments and return values.
- Similarly, alpha is not used to calculate any of the phi values and is only modified from the input value for one population. That function is trivial (1 + alpha), does not depend on phi, fsatellite or fenvironment, and is a result of a very specific modelling assumption (quenching rate proportional to mass). I would at least move the alpha calculation to a separate function, and probably suggest removing it entirely.
- This leaves just the calculation of the phi values. The active_parameters tuple can be replaced with just phi since there is no longer any dependence on mstar or alpha. I would also split into four functions: phi_centrals, phi_satellites, phi_mass_quenched and phi_satellite_quenched. This makes it much clearer from the name what each function returns and saves the user having to check the dictionary keys in the documentation when calling it. Individual functions can also be used more easily in situations where the user doesn't want to model all four populations.

skypy/galaxies/stellar_mass.py
```diff
@@ -74,3 +78,131 @@ def schechter_smf_mass(redshift, alpha, m_star, m_min, m_max, size=None,
...
+        Amplitude of the Schechter mass function.
+
+    References
+    ----------
+    .. [1] de la Bella et al. 2021, Quenching and Galaxy Demographics,
+           arXiv 2112.11110.
+
+    '''
+
+    return phi_centrals + phi_satellites
```
rrjbca, line 178: Do we definitely need such a simple function?

skypy/galaxies/stellar_mass.py
```diff
@@ -74,3 +78,131 @@ def schechter_smf_mass(redshift, alpha, m_star, m_min, m_max, size=None,
...
+        Schechter mass function amplitude of the satellite
+        active galaxies.
+
+    fenvironment: float
+        Fraction of satellite-quenched galaxies between 0 and 1.
+
+    Returns
+    -------
+    amplitude: float, (nt, nm) array_like
+        Amplitude of the Schechter mass function.
```
rrjbca, line 199:
```suggestion
    phi_satellites: array_like
        Schechter mass function amplitude of the satellite
        active galaxies.

    fenvironment: array_like
        Fraction of satellite-quenched galaxies between 0 and 1.

    Returns
    -------
    amplitude: array_like
        Amplitude of the Schechter mass function.
```
Surely `fenvironment` could also be an array here?

skypy/galaxies/stellar_mass.py
```diff
@@ -74,3 +78,131 @@ def schechter_smf_mass(redshift, alpha, m_star, m_min, m_max, size=None,
...
+           arXiv 2112.11110.
+
+    '''
+
+    if np.ndim(phi_blue_total) == 1 and np.ndim(fsatellite) == 1:
+        phi_blue_total = phi_blue_total[:, np.newaxis]
+
+    sum_phics = (1 - fsatellite) * (1 - np.log(1 - fsatellite))
+
+    return (1 - fsatellite) * phi_blue_total / sum_phics
```
rrjbca, line 116:
```suggestion
    return phi_blue_total / (1 - np.log(1 - fsatellite))
```

skypy/galaxies/stellar_mass.py
```diff
@@ -74,3 +78,131 @@ def schechter_smf_mass(redshift, alpha, m_star, m_min, m_max, size=None,
...
+        Amplitude of the Schechter mass function.
+
+    References
+    ----------
+    .. [1] de la Bella et al. 2021, Quenching and Galaxy Demographics,
+           arXiv 2112.11110.
+
+    '''
+
+    return - np.log(1 - fenvironment) * phi_satellites
```
rrjbca, line 208: This is the same as `schechter_smf_amplitude_satellites` so perhaps we can get rid of this function?

rrjbca (changes requested):
See comments for some possible further refactoring. The names of the functions seem very long can we think of anything shorter? E.g. perhaps `_amplitude_` -> `_phi_` ?

Regarding your unit tests you do far more tests of the output shapes than is really necessary. For each function one test with scalar inputs checking that it returns a scalar and another test with array inputs to check that the shapes broadcast together correctly. Any more doesn't really add anything.

On the other hand there aren't enough tests that the functions do numerically what you expect. You have one good example where `fsat/frho =0` and I would want to see more like this. E.g. for test_schechter_smf_amplitude_centrals If `fsatellite = 1 - np.exp(-1)` then you expect the function to return `0.5 * phi_blue_total`. I think for each function you should be able to come up with the input "f" values that scale the amplitude by 0, 0.5 and 1 and test them.

skypy/galaxies/stellar_mass.py
```diff
@@ -74,3 +78,131 @@ def schechter_smf_mass(redshift, alpha, m_star, m_min, m_max, size=None,
...
+        sample of galaxies, :math:`(\phi_b, \alpha_b, m_{*})`.
+
+    fsatellite: float, (nm, ) array_like
+        Fraction of active satellite galaxies between 0 and 1.
+        It could be a float or an array, depending on the model you choose.
+
+    Returns
+    -------
+    amplitude: float, (nt, nm) array_like
+        Amplitude of the Schechter mass function.
```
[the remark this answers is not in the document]
Lucia-Fonseca: @rrjbca  It might be the case where phi will have a time and a mass dependence. `phi_blue_total` could be a scalar but it can also be an array of size `nt`. Likewise, `fsatellite` can be a scalar or an array of size `nm`.  If `nt != nm` the product cannot be broadcasted.

skypy/galaxies/stellar_mass.py
```diff
@@ -74,3 +78,131 @@ def schechter_smf_mass(redshift, alpha, m_star, m_min, m_max, size=None,
...
+        active galaxies.
+
+    fsatellite: float, (nm, ) array_like
+        Fraction of active satellite galaxies between 0 and 1.
+        It could be a float or an array, depending on the model you choose.
+
+    Returns
+    -------
+    amplitude: float, (nt, nm) array_like
+        Amplitude of the Schechter mass function.
```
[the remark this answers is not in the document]
Lucia-Fonseca: same as above

skypy/galaxies/stellar_mass.py
Lucia-Fonseca: Not necessarily, but I would add it for the completeness of the model.

skypy/galaxies/stellar_mass.py
Lucia-Fonseca: Could be, `fenvironment` could depend on time, although not mass.

skypy/galaxies/stellar_mass.py
Lucia-Fonseca: is it the same?

Lucia-Fonseca: it uses `phi_satellites` that one would presumably compute with `schechter_smf_amplitude_satellites` but it has a different prefactor.

skypy/galaxies/stellar_mass.py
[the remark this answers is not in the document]
rrjbca: I don't see any reason not to make the function more flexible and allow it to take any two arrays with broadcastable shapes. If you have two 1D arrays and want the result for every combination the way to broadcast them would be:
```python
>>> x = np.arange(5)
>>> y = np.arange(7)
>>> result = schechter_smf_amplitude_centrals(x, y[:, np.newaxis])
>>> result.shape
(5, 7)
```

skypy/galaxies/stellar_mass.py
rrjbca: `-log(1-f)` is equal to `log(1/(1-f))` so both are functionally identical

Lucia-Fonseca:
@rrjbca do I need to make `hypothesis` a new dependency of skypy to run the tests?
Also I'm a bit confused about the merge conflict.
Thanks for your help.

Lucia-Fonseca:
Hi @itrharrison I can see your last commits but I do not understand why they should be part of this PR?

itrharrison:
@Lucia-Fonseca they are fixes to failing codestyle tests which had appeared because (I think) a newer version of flake8 being more strict about whitespace next to keywords like `assert`. They are indeed nothing to do with this PR but I fixed them so that the tests ran okay.

Lucia-Fonseca:
Yes, I'd wait for the rest of their approvals. 
Nonetheless, I think if a new flake8 version is making some tests or code fail but not this one on this PR, a new independent issue should be open, along with a different PR. Unless I am mistaken, @rrjbca ?

itrharrison:
@Lucia-Fonseca see #572 :-)

itrharrison:
@Lucia-Fonseca I think the only thing missing to fulfill @rrjbca's review requests is adding corner cases for the `schechter_smf_phi_mass_quenched` function. I think these testsdo that by the letter of the law, at least 😝 
```
assert phi0 == 0.5 * schechter_smf_phi_mass_quenched(phi0, phi0)
assert 0.0 == schechter_smf_phi_mass_quenched(phi0, -phi0)
assert phi0 == 1. * schechter_smf_phi_mass_quenched(phi0, 0.0)
```

Lucia-Fonseca:
@itrharrison I just submitted these changes.

> @Lucia-Fonseca I think the only thing missing to fulfill @rrjbca's review requests is adding corner cases for the `schechter_smf_phi_mass_quenched` function. I think these tests do that by the letter of the law, at least 😝
> 
> ```
> assert phi0 == 0.5 * schechter_smf_phi_mass_quenched(phi0, phi0)
> assert 0.0 == schechter_smf_phi_mass_quenched(phi0, -phi0)
> assert phi0 == 1. * schechter_smf_phi_mass_quenched(phi0, 0.0)
> ```
> 
> If you do these I will be happy to dismiss Richard's review if he doesn't respond soon himself 😃

PR #545, opened 2022-05-20
TST: assert photometric error is numerically close to the analytic value

rrjbca:
## Description

The compatibility workflow is currently failing, see [HERE](https://github.com/skypyproject/skypy/runs/6438716098?check_suite_focus=true#step:6:116)

The development version of numpy appears to cause a small numerical change in the calculation of the Rykoff photometric error. This causes the unit test to fail because it asserts that the calculated error is exactly equal to the analytic value. This PR changes the test to assert that the calculated error is numerically close to the analytic value.

Compatibility tests are shown to work with this fix [HERE](https://github.com/rrjbca/skypy/actions/runs/2357233927)

## Checklist
- [ ] Write unit tests


ntessore (approved):
Weird that whatever change happened made the situation worse. I assume (hope!) that the change is in `error` on the left-hand side, not the right-hand side?

PR #549, opened 2022-06-20
ENH: Adding model for Dust extincted ellipticity projection

### LINKED ISSUE #341: Ellipticity projection with dust absorption
## Description
Absorption produced by dust affects flux measurements as galaxies seen edge-on can potentially have significant extinction and fall outside of selection cuts. This induces a selection bias that can affect the distribution of ellipticities as the extinction increases monotonically with inclination angle.
The resultant effect can be modeled by an extincted Luminosity Function  

formula

which can be used to compute the ratio of observed galaxies as a function of inclination angle

formula
where the difference values describe observable specific conditions and selections.

## Inputs
 - Parameters for a 3D ellipsoid distribution
 - Extinction model for the observed bands
or
 - Input luminosity function

## Outputs
 - Output can be either a modified luminosity function for more general applications or a set of galaxy ellipticities drawn from a corrected distribution.

## References
Padilla & Strauss (2008)
https://ui.adsabs.harvard.edu/abs/2008MNRAS.388.1321P/abstract

jucordero:
## Description
Implements an achromatic model for dust obscured ellipticity distributions, from Padilla & Strauss 2008.
This would close #341 

The model follows a similar structure to `ryden04_ellipticity`.  The `dust_extincted_ellipticity` samples axis ratios and projects the resultant 3D ellipsoid using Binney formulation in `triaxial_axis_ratio_extincted.` Here, instead of sampling from a uniform cosine distribution for the inclination angle, `extincted_angle_schechter` is called to sample a random orientation assuming a Schechter luminosity function and the dust model.

Luminosity function parameters are assumed fixed (i.e. not arrays), which simplifies the broadcasting to get the sample size right, and also fits a more realistic use case.

## Checklist
- [x] Write unit tests

### DIFF
--- a/.zenodo.json
+++ b/.zenodo.json
@@ -44,6 +44,11 @@
             "name": "Juan Pablo Cordero"
         },
         {
+            "orcid": "0000-0003-1560-7959",
+            "affiliation": "University of Portsmouth",
+            "name": "Fox Davidson"
+        },
+        {
             "orcid": "0000-0002-8218-563X",
             "affiliation": "University of Portsmouth",
             "name": "Ginevra Favole"
--- a/docs/galaxies.rst
+++ b/docs/galaxies.rst
@@ -91,6 +91,18 @@ The following models are found in the `skypy.galaxies.stellar_mass` package.
   schechter_smf_mass
 
 
+Velocity dispersion
+-------------------
+
+The following models are found in the `skypy.galaxies.velocity_dispersion` package.
+
+.. currentmodule:: skypy.galaxies.velocity_dispersion
+.. autosummary::
+  :nosignatures:
+
+  schechter_vdf
+
+
 Reference/API
 =============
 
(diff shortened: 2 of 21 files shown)


jucordero:
I've added a notebook with some detailed explanation of how the functions are built in this enhancement to the projected-ellipticity repository (private)

https://github.com/skypyproject/projected-ellipticities/blob/master/Dust-extincted%20ellipticity%20model.ipynb

I'm not sure the unit tests I added are completely appropriate. Suggestions appreciated.

jucordero:
Just had a chat at the SkyPy meeting. In essencem, this PR models ellipticity as a function of extinction, while what we really want is a model for extinction as a function of ellipticity, which is a much simpler task.
We've also discussed about the possibility that extinction as a function of ellipticity could interfere with the k-correction templates.

PR #570, opened 2022-08-03
Test power spectra against "truth", both massless and massive nu

### LINKED ISSUE #567: TST: Improve power spectrum testing
The current test of `classy` power spectra fails, due to the interpolation creating >% level errors with respect to the previous function implementation, depending on the redshift.

We need to decide how to define a 'truth' power spectrum in order to create test data to validate which, if either of these is more correct.

This would probably require:
- [ ] creating a baseline P(k,z) where we had consistency at the 10^-4 level between camb and class run natively
- [ ]  using this P(k,z) as the test data for our implementations which include interpolation

itrharrison:
## Description

This PR addresses #567 to improve power spectrum tests. I have made the following changes to the module:

- Added `tau` and `YHe` as parameters to fully specify base cosmology consistently between CLASS and CAMB. Note that currently only `tau` is accesible.
- Added ability to explicitly specify neutrino masses. Currently in CAMB it is assumed that all the mass is in one neutrino, whilst CLASS can also deal with e.g. degenerate masses. This can easily be changed to be the same for both.
- Clarified the Hubble units in the CAMB wrapper.

And implemented new / changed the tests such that both CLASS and CAMB are tested against a 'truth' cosmology which is calculated from CAMB, with separate tests for massive and massless neutrinos. We could do a lot of playing with accuracy settings here, both in terms of the rtol in the test and the accuracy settings within each Boltzmann code. Currently these are **_specified so that the tests pass with the default accuracy settings for each Boltzmann code_**. Thoughts welcome.

Merging will close #567.

## Checklist
- [X] Follow the [Contributor Guidelines](https://github.com/skypyproject/skypy/blob/main/CONTRIBUTING.rst)
- [X] Write unit tests

### DIFF
--- a/skypy/power_spectrum/_camb.py
+++ b/skypy/power_spectrum/_camb.py
@@ -14,7 +14,7 @@ class CAMB(TabulatedPowerSpectrum):
     redshift and wavenumber using CAMB.
     '''
 
-    def __init__(self, kmax, redshift, cosmology, A_s, n_s, **kwargs):
+    def __init__(self, kmax, redshift, cosmology, A_s, n_s, tau, **kwargs):
 
         try:
             from camb import CAMBparams, model, get_results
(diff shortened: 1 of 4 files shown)


itrharrison:
Here are the two rtol plots for CAMB and CLASS for both the massless and massive (m_nu=[0.0, 0.0, 0.06]) cases.

camb-numodels
class-numodels

PR #578, opened 2022-11-17
BUG: Overgeneration of galaxies

### LINKED ISSUE #576: BUG: Overgeneration of galaxies
**Describe the bug**
When calculating the number density of galaxies for redshift sampling in 'schechter lf redshift' (/galaxies/redshift.py) there is a missing factor of h_0^3. This is required due to the units of phi_*. Not including this factor can overgenerate the number of galaxies by a factor of 3 (given h=0.7) for the entire magnitude/luminosity function. 

**To Reproduce**
Steps to reproduce the behaviour:
1. Generate a catalogue of galaxies
2. Integrate the expected magnitude Schechter function with the values used in the YAML file by volume defined by skyarea
3. Create a histogram (or extract values) from the catalogue with the same bins as the above integral
4. Plot these together, there should be a significantly increased number in the catalogue values compared to the analytical values
Please see the Jupyter Notebook and YAML files attached (I couldn't attach them in their native forms so hopefully they are still helpful) 

**Expected behaviour**
The catalogue values should exactly follow the analytical plot as they are sampled from the same distribution

**Desktop (please complete the following information):**
 - OS: Windows/WSL
 - Version 10/Debian?
[red_galaxy_text.txt](https://github.com/skypyproject/skypy/files/9992625/red_galaxy_text.txt)
[blue_galaxy_text.txt](https://github.com/skypyproject/skypy/files/9992626/blue_galaxy_text.txt)
[Galaxy_Overgeneration_(Git).pdf](https://github.com/skypyproject/skypy/files/9992627/Galaxy_Overgeneration_.Git.pdf)

Fox-Davidson:
## Description
This fixes the overgeneration of galaxies caused by the lack of a $h_0^3$ factor in the number density calculation (required due to the units of $phi_*$. Attached are a PDF of a notebook showing this result and text files for the generation of the galaxies.
This will close #576 . 

## Checklist
- [x] Write unit tests
[red_galaxy_text.txt](https://github.com/skypyproject/skypy/files/10031542/red_galaxy_text.txt)
[blue_galaxy_text.txt](https://github.com/skypyproject/skypy/files/10031543/blue_galaxy_text.txt)
[Galaxy_Overgeneration_(Git).pdf](https://github.com/skypyproject/skypy/files/10031544/Galaxy_Overgeneration_.Git.pdf)

### DIFF
--- a/skypy/galaxies/tests/test_redshift.py
+++ b/skypy/galaxies/tests/test_redshift.py
@@ -32,7 +32,7 @@ def test_schechter_lf_redshift():
     x_min = 10.**(-0.4*(M_lim - M_star))
 
     # density with factor from upper incomplete gamma function
-    density = phi_star*gamma(alpha+1)*gammaincc(alpha+1, x_min)
+    density = phi_star*gamma(alpha+1)*gammaincc(alpha+1, x_min)*np.power(cosmo.h, 3)
 
     # turn into galaxies/surface area
     density *= (sky_area * cosmo.differential_comoving_volume(z)).to_value('Mpc3')
--- a/skypy/galaxies/redshift.py
+++ b/skypy/galaxies/redshift.py
@@ -126,7 +126,7 @@ def schechter_lf_redshift(redshift, M_star, phi_star, alpha, m_lim, sky_area,
         gam[i], _ = scipy.integrate.quad(f, lnxmin[i], np.inf, args=(alpha[i],))
 
     # comoving number density is normalisation times upper incomplete gamma
-    density = phi_star*gam
+    density = phi_star*gam*np.power(cosmology.h, 3)
 
     # sample redshifts from the comoving density
     return redshifts_from_comoving_density(redshift=redshift, density=density,


skypy/galaxies/tests/test_redshift.py
```diff
@@ -32,7 +32,7 @@ def test_schechter_lf_redshift():
     x_min = 10.**(-0.4*(M_lim - M_star))
 
     # density with factor from upper incomplete gamma function
-    density = phi_star*gamma(alpha+1)*gammaincc(alpha+1, x_min)
+    density = phi_star*gamma(alpha+1)*gammaincc(alpha+1, x_min)*np.power(cosmo.h, 3)
```
philipp128, line 35: I don't think this is the right place to account for `h`. The parameter `phi_star` has the same dimension [unit] as the luminosity function, which is 1/V/luminosity [commonly Mpc^-3 mag^-1]. That means  `phi_star` must agree with these dimensions/units.

In your example, you are using the parameters of Tortorelli2020. In this work, they are giving `phi_star` in units of h_70 = H_0/70. So the input parameter must be corrected by this unit. However, as your defined cosmology includes H_0 = 70, h_70 is equal to 1. Therefore, the number that you have in your config files corresponds to units of Mpc^-3 mag^-1, which are as I said the physical units.

If we, however, want to enable `phi_star` to be given in units of h_0, we would also need to specify this in the documentation. But in this case, the parameter in your config file has the wrong value.

philipp128 (changes requested):

philipp128:
The current implementation is correct but we will add clarification about the units in the docs with a new PR. Issue #576 will be changed accordingly.

PR #630, opened 2024-09-09
Update halos module with main compatability updates.

itrharrison:
## Description

In Fox's PR #614 the tests are failing on tests not touched by him. This is (at least) because of classy failing. This PR is to update the halos module with the maintenance / compatability updates from main to start fixing such issues.

PR #632, opened 2024-09-09
MAINT: scipy deprecation fixes for halos module (attempt 2)

itrharrison:
## Description

Re-PR of #631 but not using the `modules/` name so as not to have branch protection on the source PR.


rrjbca:
See https://github.com/skypyproject/skypy/pull/630#issuecomment-2616130614

PR #643, opened 2025-01-27
MAINT: Utility function to check if a value is a pole of the gamma function

rrjbca:
## Description

The SkyPy implementation of the upper incomplete gamma function uses `scipy.special.gammasgn` to identify poles of the gamma function. However since SciPy 1.15 the behaviour of that function has changed :

```python
# SciPy 1.14.1
>>> scipy.special.gammasgn([-2, -1, 0])
array([0., 0., 0.])

# SciPy 1.15.0
>>> scipy.special.gammasgn([-2, -1, 0])
array([nan, nan,  1.])
```

This MR introduces a utility function `_is_gamma_pole` that correctly identifies poles of the gamma function using `gammasgn` for all supported versions of SciPy

## Checklist
- [x] Write unit tests

### DIFF
--- a/skypy/utils/special.py
+++ b/skypy/utils/special.py
@@ -10,8 +10,20 @@ This module computes useful special functions.
 
 """
 
+from importlib.metadata import version
+from packaging.version import parse
+
 import numpy as np
 import scipy.special as sc
+
+
+def _is_gamma_pole(a):
+    '''Utility function that returns True if a is a pole of the gamma function'''
+    s = sc.gammasgn(a)
+    if parse(version('scipy')) < parse('1.15'):
+        return s == 0
+    else:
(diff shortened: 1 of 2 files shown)


ntessore:
Is this not actually a bug in `_gammaincc`? For `x = 0` I would now naively expect

```math
Q(a, 0) = \begin{cases}
  1 & a > 0 \\
  0 & a = 0 \\
  \infty & a < 0
\end{cases}
```

but I cannot remember where we needed the special case here.

rrjbca:
Physically all distributions with x=0 and a<=0 would imply infinite source density right? So practically they are of no use to us. I don't remember if there was some mathematical reason to handle integer and non-integer a differently, I just went with the minimal change to maintain the old behaviour and unit tests with all versions of SciPy.

